PR #1, opened 2026-03-18
Setup Configuration via VS Code Tasks and CBSA base

dennis-behm:
This adds a set of VS Code tasks, that help to init a sandbox structure on z/OS Unix System Services and simulate a pipeline via a generated bash script with `clone`, `build`, `package` steps, along with the base CBSA application.

In addition, it contains a full DBB zBuilder configuration in .setup that is transferred as part of the setup task phase.


MikeFultonDev:
I would like to organize the code by component in some way but we can do that after the initial 'drop', so I approve.

PR #8, opened 2026-03-20
Enhance setup scripts and README.md

jbyibm:
This PR request has the following enhancements:
- Update README.md with some more informations
- Make user generic IBMUSER
- Set z/OS with current VSI images

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/.project
+/.settings/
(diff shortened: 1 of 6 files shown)


MikeFultonDev (approved):
Thank you - LGTM.
I don't know if it works - haven't got that far yet - but ideally if we could make the default not be /u/ibmuser/... but instead ~/ then it could be a 'reasonable' default for folks - but we can do that later.

PR #13, opened 2026-03-23
Make the pipeline script more generic

jbyibm:
Make the pipeline script more generic


MikeFultonDev (approved):
LGTM... I see .vscode/tasks.json has a '/u/ibmuser' in it - is that required? Should we force people to provide an id?

jbyibm:
Default sandbox removed

PR #20, opened 2026-03-27
CICS Db2 updates and CICS region config

AndrewTwydell:
This branch includes:
1. Migration of CBSA Customer data storage from VSAM to Db2
2. Adds zconfig configuration for a SMSSJ CICS region including CBSA
3. GRUB script filled out to run relevant existing scripts - sets up env, runs DBB, runs zconfig. Includes rearranging existing scripts to be modular, so Grub can run only the methods it needs when all files are already present.
5. JCL to create and populate Db2 database - needs migrating to more modern tooling


jbyibm (approved):
My Code is Fine

PR #23, opened 2026-04-15
Adding base CBSA zosconnect api

davidjwriter:
This is the base zosconnect api working with the CBSA application. Includes:

- openapi.yaml spec that covers 200 responses mainly
- operation and zosasset mappings -> just note that the CICS transaction/program ids might need to be changed depending on the system config
- server.xml with settings I needed to get it all working -> again, some variables would need to be externalized to polish up like the IPIC port, etc
- testing script that runs through all the different apis and verifies they are working properly
- README that tries to cover all these notes in more detail to help with migrating this into Bank of Z

The goal here is mainly to get the CBSA zosconnect work I've done over to Bank of Z so we don't have to repeat work if possible, but I put it in a separate folder for now so it didn't override any work already being done in the api folder.


MikeFultonDev:
is this ready for review? If so, can you add one or more reviewers?

AndrewTwydell:
Closing in favour of #27 which this very much influenced. Thanks @davidjwriter !

PR #27, opened 2026-04-21
Start of z/OS connect APIs with front end

AndrewTwydell:
I am building on this as we speak, but a start for adding some of the APIs. The front end also drives these already - i'm just starting it in docker alongside the z/OS connect API - although not much data is filled in yet. That's WIP!


MikeFultonDev (commented):
There are a lot of files that I 'assume' are generated since they are in folders called 'gen'. Is the intent that we will eventually (soon) be generating these files as part of our build process rather than having to check them in to git or is that incorrect?

AndrewTwydell:
Thanks @MikeFultonDev - I believe @stewartfrancis asked Andrew S the same question! I'm not sure which files should be checked-in vs generated at build time currently

AndrewTwydell:
@stewartfrancis I've removed the unused Copybook change. Plenty of copybook and cobol changes inbound after I've adapted the data structures to better accommodate the openapi spec!! 💪🏼

stewartfrancis (approved):

PR #29, opened 2026-04-29
Add customer email to Bank of Z CICS

stewartfrancis:
This contains Ishan's changes plus some of my review changes


AndrewTwydell (approved):
I've tested this and it works great. Only suggestion would be to generate the email addresses based off the customer name instead of the ID but that can be done at a later date!

PR #30, opened 2026-04-30
Split customer name and address into individual fields

AndrewTwydell:
Changes here split a few existing customer fields (name, address, DOB, etc) into individual elements (title, first name, last name, address line 1, postcode, city, etc). This allows better mapping to the modern openapi format.

Screenshot 2026-05-11 at 6 03 13 pm

Screenshot 2026-05-11 at 6 03 28 pm


.setup/jcl/Db2-drop.jcl
```diff
@@ -13,11 +13,23 @@
...
+DROP INDEX IBMUSER.ACCTINDX;
+DROP TABLE IBMUSER.ACCOUNT;
+
 DROP DATABASE BANKZ;
 
-DROP STOGROUP ACCOUNT;
-DROP STOGROUP PROCTRAN;
-DROP STOGROUP CONTROL;
 DROP STOGROUP CUSTOMER;
-DROP ALIAS IBMUSER.CUSTOMER;
```
stewartfrancis, line 22: Do we not need to drop this alias any more?  Not sure what it does!

src/base/cics/cobol/BANKDATA.cbl
```diff
@@ -543,16 +557,29 @@
...
-
-                   COMPUTE CUSTOMER-BIRTH-DAY =((28 - 1)
+                      INTO CUSTOMER-ADDR-LINE1 OF CUSTOMER-ADDRESS
+                   
+                   MOVE TOWN(TOWN-PTR) TO CUSTOMER-CITY
+                      OF CUSTOMER-ADDRESS
+                   
+      *            Generate random UK-style postcode (e.g., SW1A 1AA)
+                   COMPUTE WS-POSTCODE-NUM =
+                      ((9999 - 1000) * FUNCTION RANDOM) + 1000
```
stewartfrancis: Is there some magic going on here, or is 9999-1000 not always 8999?

stewartfrancis: I guess this is to kind of show the range of random number that's being generated.  Seems odd to me, but it's the existing style in this module

src/base/cics/cobol/BANKDATA.cbl
```diff
@@ -543,16 +557,29 @@
...
+                   STRING 'SW' DELIMITED BY SIZE
+                          WS-POSTCODE-NUM DELIMITED BY SIZE
```
stewartfrancis, line 569: Doesn't this mean we get a SW1234 1AA postcode?  Ah yes it does, I can see it in your screenshot!

src/base/cics/cobol/BNK1CAC.cbl
```diff
@@ -156,7 +156,7 @@
           03 COMM-INTRT             PIC 9(4)V99.
           03 COMM-OVERDR            PIC 9(8).
 
-       PROCEDURE DIVISION USING DFHCOMMAREA.
```
stewartfrancis, line 159: Do we need to drop the commarea as an argument here?  Seems like we're supposed to need it?

src/base/cics/cobol/BNK1DCS.cbl
```diff
@@ -117,7 +117,7 @@
           03 COMM-ADDR-UPD3           PIC X(40).
 
        01 INQCUST-COMMAREA.
-           COPY INQCUST.
```
stewartfrancis, line 120: Do we need to change this to use the z/OS connect copybook?

src/base/cics/cobol/CRDTAGY3.cbl
```diff
@@ -42,21 +42,38 @@
...
-          03 WS-CONT-IN-DATE-OF-BIRTH   PIC 9(8).
-          03 WS-CONT-IN-DOB-GROUP REDEFINES WS-CONT-IN-DATE-OF-BIRTH.
-             05 WS-CONT-IN-BIRTH-DAY    PIC 99.
-             05 WS-CONT-IN-BIRTH-MONTH  PIC 99.
-             05 WS-CONT-IN-BIRTH-YEAR   PIC 9999.
-          03 WS-CONT-IN-CREDIT-SCORE    PIC 999.
-          03 WS-CONT-IN-CS-REVIEW-DATE  PIC 9(8).
-          03 WS-CONT-IN-SUCCESS         PIC X.
-          03 WS-CONT-IN-FAIL-CODE       PIC X.
+          03 WS-CONT-CUSTOMER-RECORD.
```
stewartfrancis: I bet there's a way we can reuse these structures

src/base/cics/cobol/BNK1DCS.cbl
```diff
@@ -331,7 +364,7 @@
            EXEC CICS
                 RETURN TRANSID('ODCS')
                 COMMAREA(WS-COMM-AREA)
-                LENGTH(266)
+                LENGTH(504)
```
stewartfrancis: Is this the length of the copybook?  Can we calculate it dynamically as you've done elsewhere?

src/base/cics/cobol/CRECUST.cbl
```diff
@@ -1497,22 +1574,33 @@
       *
       *    Ensure that the Date Of Birth is valid
       *
-           IF COMM-BIRTH-YEAR < 1601
+           DISPLAY '7A - Starting DATE-OF-BIRTH-CHECK'
```
stewartfrancis: Looks like you left some debugging statements in?

stewartfrancis (changes requested):
Quite a few whitespace only changes too.  Not a problem if they're making the whitespace usage canonical, so we don't see more changes in the future.

src/base/cics/cobol/BANKDATA.cbl
AndrewTwydell: Indeed, it's creating the random number, and follows existing ways of doing it as you said! 😄

src/base/cics/cobol/BNK1CAC.cbl
AndrewTwydell: We don't need it here because it's provided in the LINKAGE SECTION, passed in by CICS.

.setup/jcl/Db2-drop.jcl
AndrewTwydell: Checked and confirmed - alias references no longer exist!

src/base/cics/cobol/CRECUST.cbl
AndrewTwydell: Removed!

src/base/cics/cobol/BNK1DCS.cbl
AndrewTwydell: Yes indeed!

src/base/cics/cobol/CRDTAGY3.cbl
AndrewTwydell: I've reused the CUSTOMER copybook in all the credit check programs. Definitely more optimisation we can do regarding reused down the line

src/base/cics/cobol/BNK1DCS.cbl
AndrewTwydell: Need, no, can and have, yes. Makes sense so COBOL and ZC use the same copybook

src/base/cics/cobol/BANKDATA.cbl
AndrewTwydell: Corrected to be a valid format when generated. Currently anything can be specifed when creating a customer so perhaps some validation to add down the line

AndrewTwydell: Screenshot 2026-05-18 at 10 58 38 am

stewartfrancis (approved):

PR #36, opened 2026-05-13
Add CICS account transaction history

AndrewTwydell:
This adds account transaction history to Bank of Z for CICS programs.

- New programs added to retrieve account transactions already writing to the PROCTRAN Db2 table.
- z/OS connect APIs filled out to get all transactions for a specified account, and details about a specific transaction
- Added pages to GUI to display this info


MikeFultonDev:
@AndrewTwydell is it safe to delete this branch? If do, please do so

PR #37, opened 2026-05-13
Update pipeline setup scripts

jbyibm:


.setup/create/create-application.sh
```diff
@@ -0,0 +1,84 @@
...
+bash ../tasks/task-dbb-build.sh full
+
+# =========================
+# Stage 3: Deploy Build
+# =========================
+cd "$SCRIPTS_DIR"
+print_stage "STAGE 3: Deploy Build"
+bash ../tasks/task-wazi-deploy.sh true&
+# ZOAU Issue with ZOWE
+PID=$!
```
MikeFultonDev, line 41: why do you run in background and wait instead of just running?

.setup/create/create-taz-configuration.sh
```diff
@@ -0,0 +1,309 @@
...
+         TITLE 'DFHPLT - IBM z/OS Debugger CICS Sample PLT'
+         DFHPLT TYPE=INITIAL
+         DFHPLT TYPE=ENTRY,PROGRAM=DFHDELIM
+         DFHPLT TYPE=ENTRY,PROGRAM=EZACIC20
+         DFHPLT TYPE=ENTRY,PROGRAM=EQA0CPLT
+         DFHPLT TYPE=FINAL
+         END DFHPLT
+EOF
+
+iconv -f ISO8859-1 -t IBM-1047 "$tmp_ascii" > "$tmp_ebcdic"
```
MikeFultonDev: since we are already using Z Open Automation Utilities, you can use a2e here instead of iconv - it will also set code pages which is nice.

.setup/create/create-taz-configuration.sh
```diff
@@ -0,0 +1,309 @@
...
+PLTPI=PI
+TCPIP=YES
+/*
+EOF
+
+# =========================
+# Convert PROC to EBCDIC and copy to SYS1.PROCLIB
+# =========================
+if ! iconv -f ISO8859-1 -t IBM-1047 "$PROC_TMP" > "$PROC_TMP.iconv"; then
+    print_error "${RED}[TAZ-INSTALL]${NC} iconv to EBCDIC failed"
```
MikeFultonDev: use a2e

.setup/lib/prerequisites.sh
```diff
@@ -21,7 +21,7 @@ check_dbb_installation() {
 # Function to check if Java is installed and valid
 # Usage: check_java_installation
 check_java_installation() {
-    if [ -z "$JAVA_HOME" ]; then
+    if [ -z "${JAVA_HOME:-}" ]; then
```
MikeFultonDev, line 24: Be consistent with dbb and Java - probably nice to default them both to reasonable values. JAVA_HOME doesn't seem to have a default here.

.setup/lib/utilities.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/env bash
+
+# Function to parse YAML config - reads a key within a specific section
+# Usage: get_section_value <section> <key>
+
+
+_get_section_value_() {
```
MikeFultonDev, line 7: zopen has a 'yq' utility. We might want to consider using that at some point.

.setup/tasks/task-wazi-deploy.sh
```diff
@@ -0,0 +1,190 @@
...
+# =========================
+finalize_results() {
+    RC=$?
+
+    mkdir -p "$outputDir" "$evidenceDir"
+    cd "$SCRIPTS_DIR"
+
+    if ls logs/*.log >/dev/null 2>&1; then
+        chtag -tc ISO8859-1 logs/*.log
+        iconv -f IBM-1047 -t ISO8859-1 "$outputDir/wazideploy-generate.log" > "$outputDir/wazideploy-generate.console.log"
```
MikeFultonDev: use e2a from Z Open Automation Utilities

MikeFultonDev (approved):
I have comments but overall approve - nothing major

.setup/create/create-application.sh
jbyibm: This is to bypass a ZOAU bug with ZOWE. If we don't start the process in the background, then the process will never return. The ZOAU dev team is aware of this, and they will fix it.

.setup/create/create-taz-configuration.sh
jbyibm: Go to know I'll use it.

.setup/create/create-taz-configuration.sh
jbyibm: Go to know I'll use it.

.setup/lib/prerequisites.sh
jbyibm: It would be better to use a unique Java version (version 21). But with Gradle, we still need Java 17 as it does not work with Java 21 (new encoding management with Java 21). This is needed to build z/OS Connect APIs during DBB Build.

.setup/tasks/task-wazi-deploy.sh
jbyibm: Go to know I'll use it.

.setup/lib/utilities.sh
jbyibm: Unfortunately, YQ is not in our ZDVT. It is the first thing I looked at. How can make it available ?

PR #38, opened 2026-05-18
Update CICS setup, use common script for Local and USS automation

AndrewTwydell:
Adds on @jbyibm setup changes, moving logic to 1 `-common` script, with some setup in `-local` script if running from VSCode tasks locally. 

- Running with GRUB, or manually on USS, just uses `setup-common.sh`
- Running from VS Code tasks uses `setup-local.sh` which calls `setup-common.sh` after some prereqs.


jbyibm (approved):
Too many changes to check. So I'll merge and check.

PR #44, opened 2026-05-22
Update pipeline and build so dbb builds zosconnect and wazi deploy deploys it

davidjwriter:
Reorganized the installation pipeline to separate infrastructure setup from application deployment, and integrated Wazi Deploy for automated z/OS Connect artifact deployment.

### New Order
1. **Install Middleware** (CICS, z/OS Connect, DB2 Tables)
2. **DBB Build** (all artifacts)
3. **Wazi Deploy** (CICS/DB2 + z/OS Connect)

### z/OS Connect Server Setup
- `setup-zosconnect-server.sh` now only creates the server and JCL proc
- Removed manual `api.war` deployment
- Removed manual `server.xml` and `cics.xml` copying
- Server is created empty, ready for Wazi Deploy
- **Future**: Can be replaced with zconfig for server creation

### Wazi Deploy Integration
- Added second deployment phase for z/OS Connect artifacts
- Automatically deploys `api.war`
- Backs up existing WAR files before deployment
- Starts server and refreshes configuration automatically
- Uses dynamic job name: `BAQ${APP_BASE_NAME}` (e.g., `BAQBANKZ`)

### New Files
- `.setup/deploy/zosconnect-deployment-method.yml` - Wazi Deploy method for z/OS Connect
- Updated `.setup/deploy/Development.yml` - Added z/OS Connect configuration

## Future Enhancements

- Enable `server.xml` and `cics.xml` deployment with variable substitution
- Replace `setup-zosconnect-server.sh` with zconfig integration


.setup/build/groovy/convert-encoding.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#########################################################
```
MikeFultonDev: This shouldn't be necessary if .gitattributes is set up right (treat .groovy files the same as .cbl files) correct?

MikeFultonDev: If it _is_ necessary it should read the tag first to see what code page it is in and only iconv it if it is necessary. It is better to use a2e as well if necessary, which is a utility provided by Z Open Automation Utilities that does an in-place conversion (there is also an e2a if you need it).
Also - the file should be tagged (I think) as ISO8859-1 and NOT UTF-8 since we can't tolerate a data conversion failure from muti-byte to single byte conversion. If we DO need to keep this as UTF-8, we should validate the conversion is successful.

.setup/config/config.yaml
```diff
@@ -40,7 +40,7 @@ dbb:
   # DBB installation directory
   dbb_home: /usr/lpp/IBM/dbb
   # Java runtime for DBB
-  java_home: /usr/lpp/java/java21/current_64
+  java_home: /usr/local/sandboxes/tools/J17.0_64
```
MikeFultonDev: Why are we 'dropping down' to Java 17?

.setup/config/config.yaml
```diff
@@ -126,8 +135,15 @@ fronted:
 zoau:
   zoau_home: /usr/lpp/IBM/zoautil
 
-# CICS attibutes
+# Zowe CLI configuration
+zowe:
+  # RSE profile name to use (optional, defaults to 'rse')
+  # Set via environment variable: export ZOWE_RSE_PROFILE=manzanita
```
MikeFultonDev: I don't know what manzanita is - we should use something more descrptive for open source

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -57,71 +57,63 @@ fi
...
+# Remove temp file if it exists from previous run
+rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl.ebcdic"
 
-# =========================
-# Start server
-# =========================
-opercmd "S BAQ${APP_BASE_NAME}" & 2>/dev/null
-sleep 10
-print_success "z/OS Connect server started successfully"
+# Convert to EBCDIC and pad lines to 80 characters
```
MikeFultonDev: Use a2e instead (it comes with Z Open Automation Utilities and you use 'dcp' a couple lines later)

.setup/tasks/task-wazi-deploy.sh
```diff
@@ -59,8 +60,16 @@ finalize_results() {
 
     if ls wazideploy*.log >/dev/null 2>&1; then
         chtag -tc IBM-1047 wazideploy*.log
-        a2e -f IBM-1047 -t ISO8859-1 "$outputDir/wazideploy-generate.console.log"
-        a2e -f IBM-1047 -t ISO8859-1 "$outputDir/wazideploy-deploy.console.log"
+        # Convert CICS/DB2 logs if they exist
+        [ -f "$outputDir/wazideploy-generate-cics.console.log" ] && \
+            a2e -f IBM-1047 -t ISO8859-1 "$outputDir/wazideploy-generate-cics.console.log"
```
MikeFultonDev: Use e2a and then you don't need the from and to.

MikeFultonDev (commented):
Minor comments.

.setup/config/config.yaml
jbyibm: J17.0_64 was the only version that worked before the fix in Java 21. So now we use Java 21 for all tools.

.setup/config/config.yaml
jbyibm: Changed to bankz.  But will be removed when we remove the Zowe CLI.

.setup/setup/setup-zosconnect-server.sh
jbyibm: Done

.setup/tasks/task-wazi-deploy.sh
jbyibm: Code changed

.setup/build/groovy/convert-encoding.sh
jbyibm: This file is not used. I remove it

jbyibm (approved):
I revisited the code.

PR #54, opened 2026-06-02
Separate the setup and the pipeline

jbyibm:


.setup/config/config.yaml
```diff
@@ -99,6 +99,18 @@ wazideploy:
...
   deploy_log_dir: ${sandbox.path}/logs/deploy
 
+# TAZ Unit Test configuration
+taz:
+  # TAZ CLI installation directory
+  taz_home: /usr/local/sandboxes/tools/taz-280/test-cli
+  # Root directory containing test suites
+  test_dir: ${sandbox.path}/${REPO_NAME}
+  # JES procedure library
+  proclib: SYS1.PROCLIB
```
MikeFultonDev, line 109: what is our philosophy on hardcoding these types of items? Should we have an env var for these that we use in the config file, or should we expect users to need to update a file? 
Is this the only config file people need to change the hard-coded values or are there a few? 
Want to have it so people can easily configure for their system in one place - if this is the one place, and we document it in getting started, all good!

.setup/setup-common.sh
```diff
@@ -419,15 +338,18 @@ main() {
...
             ;;
         install-bank-of-z)
-            stage_build_bank_of_z full
-            stage_deploy_bank_of_z
+            chmod +x ${SCRIPTS_DIR}/pipeline-common.sh
+            bash ${SCRIPTS_DIR}/pipeline-common.sh build-and-deploy full &
+            PID=$!
+            # Wait for deployment to complete (ZOAU/ZOWE ISSUE)
+            if wait "$PID"; then
+                print_success "Remote pipeline completed successfully"
```
MikeFultonDev, line 346: do we not need to check a return code to see if it was successful?

.setup/build/datasets.yaml
```diff
@@ -118,3 +118,7 @@ variables:
   - name: SEQASAMP
     value: EQA.V16R0M0.SEQASAMP
 
+  # IBM ADFz Common Components
```
MikeFultonDev, line 121: General comment - Do any of these changes change the externals on how to build? If not, then all good. If there are, then the getting started doc(s) should be updated too

.setup/config/config.yaml
jbyibm: We still have some work to do to have one place config. For example, DBB and Wazi Deploy also have some files. We are exploring how to merge with zconfig.

.setup/setup-common.sh
jbyibm: Yes better

.setup/config/config.yaml
MikeFultonDev: can you open a defect for this work so we don't lose track and assign it to the right person?

MikeFultonDev (approved):
Once defects opened for enhancements requested here, can merge.

.setup/setup-common.sh
MikeFultonDev: I misunderstood 'wait PID' - all good.

PR #59, opened 2026-06-03
Add cics debug config

AndrewTwydell:
Adds the debug port to config, adds libraries and SIT params to CICS region config, and overrides the CSD definition port.


jbyibm:
I'll let @stewartfrancis review - but thank for this.

.setup/zconfig/debug-definitions.yaml
dennis-behm, line 1: @AndrewTwydell - I am looking at this Debugger Config Guide, but I don't see that many program definitions mentioned in there - https://www.ibm.com/docs/en/debug-for-zos/17.0.x?topic=guide-adding-support-debugging-under-cics

Are all these required?

AndrewTwydell: That page says 
> The CICS definitions are in the EQACCSD and EQACDCT members of the hlq.SEQASAMP data set

Are these the definitions you're referring to? That's where I got them from

dennis-behm: @AndrewTwydell  - oh boy. I did not know that these are that many :)  I ignore my comment

.setup/zconfig/debug-definitions.yaml
```diff
@@ -0,0 +1,1513 @@
...
+      name: EQA00E53
+      group: EQA
+      datalocation: ANY
+      cedf: NO
+      concurrency: THREADSAFE
+      execkey: CICS
+  - program:
+      name: EQA00E62
+      group: EQA
+      datalocation: ANY
```
stewartfrancis: Datalocation: any is the default applied by CICS, so we shouldn't need all these lines

stewartfrancis (changes requested):

.setup/zconfig/debug-definitions.yaml
AndrewTwydell: All gone!

stewartfrancis (approved):

PR #63, opened 2026-06-04
Set java home before validation step

AndrewTwydell:
Getting a failure validating DBB installation in `validate-install.sh` due to JAVA_HOME not being set. This sets it prior to validation.


jbyibm (approved):
Look Good

PR #65, opened 2026-06-04
Fix frontend and config packaging to skip unchanged files in pipeline/impact builds

davidjwriter:
Fixes VanillaFrontend and ServerXmlPackager scripts to properly detect file changes in pipeline/impact builds, preventing unnecessary rebuilds when frontend or config files haven't changed.

## Problem
Frontend WAR and server/cics XML files were being rebuilt and packaged on every pipeline/impact build, even when no relevant files had changed. This occurred because the scripts were checking BUILD_LIST, which only contains buildable source files (COBOL, BMS, etc.) and doesn't include non-buildable files like HTML or XML.

## Solution
- Check `CHANGED_FILES`, `DELETED_FILES`, and `RENAMED_FILES` context variables instead of BUILD_LIST
- Use proper TaskConstants (`CHANGED_FILES`, `DELETED_FILES`, `RENAMED_FILES`, `LIFECYCLE`)
- Add informative console output showing build decisions

### VanillaFrontend.groovy
- Now checks changed/deleted/renamed files for frontend directory matches
- Skips packaging if no frontend files changed in pipeline/impact builds
- Always packages in full builds

### ServerXmlPackager.groovy
- Now checks changed/deleted/renamed files for server.xml/cics.xml matches
- Skips packaging if no config files changed in pipeline/impact builds
- Always packages in full builds


jbyibm (approved):
Look Good

Issue #56, opened 2026-06-03
Sometime allocating ${APP_BASE_NAME}.${APP_VERSION}.LOADLIB failed


henytjin:
https://github.com/IBM/Bank-of-Z/blob/1b6645bd590b1cbe48dbf7dacebc5f525578daff/.setup/setup/setup-cics-region.sh#L49

Hello, I found the command to allocate for loadlib sometimes (sporadically) failed. Somehow `drm "${APP_BASE_NAME}.${APP_VERSION}.*"` seems not clear the datasets in time. 

In zVDT system I used, I tried on to delete the library first and it seems to ensure the preceding allocation successful. 
```bash
# Delete existing dataset to avoid error when alloc
tsocmd "DELETE '${APP_BASE_NAME}.${APP_VERSION}.LOADLIB'" 2>&1 || true

tsocmd "ALLOC DA('${APP_BASE_NAME}.${APP_VERSION}.LOADLIB') NEW CATALOG DSNTYPE(LIBRARY) DSORG(PO) RECFM(U) BLKSIZE(32760) SPACE(5,5) CYL DIR(20)"
```

So maybe we can add that to ensure loadlib is created and script can continue?

jbyibm:
Proceed with a PR.  Thanks

henytjin:
To avoid conflict, I will wait till https://github.com/IBM/Bank-of-Z/pull/59 merged, and put in a PR.

henytjin:
@jbyibm Hello, I have created a PR for the minor fix: https://github.com/IBM/Bank-of-Z/pull/72. But I can’t find adding reviewer button on the button. Will you know how can I request for review for the PR?

henytjin:
This is issue should be fixed by PR https://github.com/IBM/Bank-of-Z/pull/72

PR #72, opened 2026-06-09
Ensure drm ${APP_BASE_NAME}.${APP_VERSION}.* completed before alloc

henytjin:
https://github.com/IBM/Bank-of-Z/blob/1b6645bd590b1cbe48dbf7dacebc5f525578daff/.setup/setup/setup-cics-region.sh#L49 
I have noticed in slower zVDT instance, script failed at this line. (Reported in issue: https://github.com/IBM/Bank-of-Z/issues/56) 
I suspect it might be the earlier command to `drm "${APP_BASE_NAME}.${APP_VERSION}.*"` is not totally complete yet. 
Not having it as background process will ensure completion of the command when the script proceed to allocate for the load library.

### DIFF
--- a/.setup/setup/setup-cics-region.sh
+++ b/.setup/setup/setup-cics-region.sh
@@ -40,12 +40,12 @@ export LIBPATH="$ZOAU_HOME/lib:${LIBPATH:-}"
 # Ignore errors if already cancelled
 # =========================
 set +e
-jcan P "CICS${APP_SHORT_NAME}" & 2>/dev/null
-opercmd "C CICS${APP_SHORT_NAME}" & 2>/dev/null
+jcan P "CICS${APP_SHORT_NAME}"  2>/dev/null
+opercmd "C CICS${APP_SHORT_NAME}"  2>/dev/null
 sleep 10
 drm "${APP_BASE_NAME}.${APP_VERSION}.*" 2>/dev/null
-drm "${APP_BASE_NAME}.CICS${APP_SHORT_NAME}.*" & 2>/dev/null
-drm "${APP_BASE_NAME}.DBB.*" & 2>/dev/null
+drm "${APP_BASE_NAME}.CICS${APP_SHORT_NAME}.*"  2>/dev/null
+drm "${APP_BASE_NAME}.DBB.*"  2>/dev/null
 sleep 5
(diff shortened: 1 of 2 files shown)


.setup/setup/setup-cics-region.sh
```diff
@@ -43,7 +43,7 @@ set +e
 jcan P "CICS${APP_SHORT_NAME}" & 2>/dev/null
 opercmd "C CICS${APP_SHORT_NAME}" & 2>/dev/null
 sleep 10
-drm "${APP_BASE_NAME}.${APP_VERSION}.*" & 2>/dev/null
+drm "${APP_BASE_NAME}.${APP_VERSION}.*" 2>/dev/null
 drm "${APP_BASE_NAME}.CICS${APP_SHORT_NAME}.*" & 2>/dev/null
```
jbyibm: Unfortunately, this is not possible due to a bug in ZOAU and/or ZOWE. Reason why we start the commands in the background. This task uses Zconfig that uses ZOAU. And in our scenario, ZOWE CLI can be use to lauch the setup.

What you can do is to replicate what I did for Wazi Deploy. In `setup-common.sh` replace this code:

```
    set -o pipefail
    if  .setup/setup/setup-cics-region.sh; then
        print_success "Bank of Z application setup completed successfully"
    else
        print_error "Failed to install Bank of Z"
        exit 1
    fi
```
by:
```
   set -o pipefail
   .setup/setup/setup-cics-region.sh&
    PID=$!
    # Wait for CICS setup to complete (ZOAU/ZOWE ISSUE)
    if wait "$PID"; then
        print_success "Bank of Z application setup completed successfully"
    else
        print_error "Failed to install Bank of Z"
        exit 1
    fi
```


Then in `setup-cics-region.sh` you should be able to remove all `&`.

Before pushing, you must test with ZOWE CLI the setup of the pipeline.

jbyibm (changes requested):
Needs change.

henytjin:
@jbyibm Hello Jean, I have tested on running set-cics-region.sh as background process and having all zoau command in script to run synchronously, and build-install still works fine.

jbyibm (approved):
Looks Good

PR #74, opened 2026-06-11
Created customer api

Sash9:
Added a new Create Customer API that allows the application to create customer records through the Bank of Z backend. The endpoint accepts basic customer details like name, date of birth, phone number, and address, then returns the newly created customer ID and sort code. This update also includes the supporting API configuration and service setup needed to make the endpoint available and working in the local environment.

- How It Works:

User sends customer details via POST request to /customers.
The API transforms the JSON data into a format the mainframe understands.
Calls the mainframe COBOL program (CRECUST) which:
Validates the customer information
Saves the customer to the database
Returns the new customer ID and bank sort code

- What Was Added:

REST API endpoint: POST /customers
Request/response mappings to connect modern JSON API with legacy mainframe system.
Configuration to call the CRECUST mainframe program.
Proper error handling (returns 201 on success, 500 on failure).
Customer input validation of all fields

Screenshot 2026-06-16 at 3 26 14 PM
Screenshot 2026-06-16 at 3 27 05 PM


src/api/src/main/liberty/config/server.xml
```diff
@@ -2,28 +2,4 @@
   <featureManager>
     <feature>zosconnect:zosConnect-3.0</feature>
   </featureManager>
-
```
MikeFultonDev: Why was this deleted?

src/frontend/customer-create.html
```diff
@@ -136,14 +136,20 @@ <h1>Create new customer</h1>
             }
             
             // Prepare data
-            const customerAddress = `${addressLine1}, ${city}`;
-            const customerName = `${title} ${fullName}`;
+            const nameParts = fullName.trim().split(/\s+/);
```
MikeFultonDev: Are there any tests for this? Who ensures that there is no overflow if a customer name is too long that is sent to the back end (for example)? What type of message is produced?

src/api/src/main/liberty/config/server.xml
AndrewTwydell: I suspect this was generated from the z/OS connect designer which likes this being empty? 🤔

src/frontend/customer-create.html
AndrewTwydell, line 1: I think this page should ask for all details in separate fields here (title, first name, last name, address line 1, line 2, city, postcode, country, etc) rather than doing some client-side manipulation. The COBOL / Db2 stores it like this, so we could just ask the user in the same format

AndrewTwydell: It'll also prevent any erroneous input data causing problems..!

AndrewTwydell (changes requested):
The API changes look reasonable, about to test in isolation. I think the frontend should be adapted to handle the more granular customer data

src/api/src/main/liberty/config/server.xml
Sash9: The z/OS connect designer overwrites server.xml due to the new changes in openapi.yaml. 

Suggested change: At deploy time, we can create the z/OS connect server with an empty server.xml that the designer owns. During app deployment, we move the CICS and IMS config into configDropins so it can be reloaded independently, and move api.war into the apps directory.

src/frontend/customer-create.html
Sash9: Input validation and tailored error messages are not yet fully implemented. I will add more safety precautions to handle edge cases such as field length limits.

src/frontend/customer-create.html
Sash9: Sounds good! I'll work on those changes.

AndrewTwydell:
Hey, I got this working which is great, but have a couple of questions.

1. z/OS connect designer is changing a lot of the files when I run it. Does that happen for you, and should the be committed to prevent people running into this themselves - mostly formatting changes etc.
2. A big change is designer removing this block. If I add the fields in designer, it works and formats differently. Perhaps a syntax error in the JSONata??

Your branch after starting designer
Screenshot 2026-06-17 at 10 28 33 am

The changes after I added the fields in designer
Screenshot 2026-06-17 at 10 29 42 am

AndrewTwydell (approved):

PR #76, opened 2026-06-11
Fix destination path for zosconnect_config template

jbyibm:


jbyibm:
@dennis-behm I have to revert urgently; the setup is not working anymore.

dennis-behm (approved):

PR #80, opened 2026-06-15
Hide warnings on yaml scanner

jbyibm:


dennis-behm (changes requested):
@jbyibm 
Based on the [scanner configuration](https://www.ibm.com/docs/en/adffz/dbb/3.0.x?topic=index-task-scannerinit), we will need to expand the `excludeFileList` and add `**/*.yaml` to the list:
```
[".*","**/.*","**.yaml", "**/*.yaml","**.groovy","**.json","**.md"]
```

jbyibm:
@dennis-behm have a better solution

PR #84, opened 2026-06-16
Fix: Use SANDBOX_DIR variable instead of hard-coded path for WLP_USER…

adn1107:
…_DIR in JCL

The WLP_USER_DIR in the z/OS Connect JCL proc was hard-coded to /usr/local/sandboxes/bank-of-z/zosconnect-server instead of using the configurable SANDBOX_DIR variable from config.yaml.

This change makes the setup portable across different environments by using ${SANDBOX_DIR}/zosconnect-server which resolves to the sandbox.path value defined in the configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ zowe.config.user.json
 vsix-extensions/
 *.vsix
 /.taz-unit-test
+tools/
+imsbank_provisioning/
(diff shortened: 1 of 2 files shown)


jbyibm (approved):
Looks Good

PR #85, opened 2026-06-16
Fix: Enable shell variable expansion in JCL generation for WLP_USER_DIR

adn1107:
Changed heredoc delimiter from << 'EOF' to << EOF to allow shell variable expansion. This ensures ${SANDBOX_DIR} is replaced with the actual sandbox path value when generating the JCL proc.

Without this fix, the literal string ${SANDBOX_DIR} was written to the JCL, causing z/OS Connect to fail with 'server does not exist' error because it couldn't resolve the path.

Now the JCL will contain the actual path from config.yaml, making the setup portable across different environments.


QihanQiu1 (approved):
Looks Good

PR #89, opened 2026-06-19
Direct YAML files to Dummy scanner 

dennis-behm:
This configures the build to only look for files in the `src` folders and configure the dbb scanners for YAML


AndrewTwydell (approved):
The build was successful for me

PR #97, opened 2026-06-23
Remove the node.js development server

stewartfrancis:
Remove the node.js development server used for testing the z/OS connect APIs in designer, and replace with nginx, so we don't have to maintain a separate server


AndrewTwydell (approved):
Thanks Stew, testing now but looks good! Will merge shortly

AndrewTwydell:
@stewartfrancis when running it locally, the frontend was still requesting `http://localhost:3001/api/customers/1234`, we actually want to request `http://localhost:9080/customers/1234` (no /api). 

I don't know if the location block in nginx is supposed to resolve this? 

Bob advised this instead which worked when running locally, but I don't know if you had wider reasons for your nginx conf
```
    # Proxy API requests to the backend zosConnect service
    # Frontend makes requests to /api/* which nginx forwards to zosConnect:9080/*
    location /api/ {
        proxy_pass http://zosConnect:9080/;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
    }
```

AndrewTwydell:
I have not forgotten about this, I will rebase and double check shortly what's already gone in and what's left to remove node 👍🏼

MikeFultonDev:
@AndrewTwydell is this your item? If so, can you assign it to yourself? At which point presumably @stewartfrancis should be the reviewer? Can we get this closed off?

AndrewTwydell:
Finally got round to fixing the last couple of bits - @stewartfrancis a onceover from you would be great as it was your work originally!

PR #109, opened 2026-06-24
Source user's .profile in setenv.sh

jbyibm:
Added sourcing of the user's .profile to setenv.sh  as source /etc/profile does not source user's .profile in z/OS


adn1107 (approved):
ok

PR #120, opened 2026-06-26
Intelligent incremental build and deploy via GRUB (DO NOT MERGE)

QihanQiu1:
Adds git-based change detection to build and deploy only what actually
changed, rather than performing full rebuilds every time.

Changes:
- .grub/grub-incremental-build-and-deploy.sh: Entry point script that
  uses grub_client to sync and execute the orchestrator remotely
- .grub/scripts/build-and-deploy-orchestrator.sh: Remote script that
  runs DBB in pipeline mode and selects minimum Wazideploy deployment
  tags based on what type of files changed (frontend vs backend vs mixed)
- .setup/build/groovy/VanillaFrontend.groovy: Added git diff fallback
  to detect HTML/CSS/JS changes independently of DBB's scanner (which
  excludes those file types)
- dbb-app.yaml: Exclude html/css/js from ImpactAnalysis scanner to
  prevent WARNING build status blocking DBB baseline advancement"

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,8 +8,11 @@
 *.rex zos-working-tree-encoding=ibm-1047 
 #.txt zos-working-tree-encoding=ibm-1047 
 *.groovy zos-working-tree-encoding=ibm-1047 
-*.sh zos-working-tree-encoding=ibm-1047 
-*.properties zos-working-tree-encoding=ibm-1047 
+*.sh zos-working-tree-encoding=ibm-1047
+# GRUB wrapper scripts should stay UTF-8 (contain emojis, run locally)
+.grub/*.sh zos-working-tree-encoding=utf-8
+.grub/**/*.sh zos-working-tree-encoding=utf-8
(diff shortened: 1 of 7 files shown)


jbyibm (changes requested):
Please mark this as a draft. I don't understand at all the goal of this PR.

jbyibm:
@QihanQiu1   Can you connect with @davidjwriter if you need to update DBB scripts?

As I said, DBB already has the possibility to do impact builds and can also compute dependencies. For this, you need to use the `pipeline-remote.sh` script.

@MikeFultonDev  FYI

MikeFultonDev:
@jbyibm @QihanQiu1 @davidjwriter Please review the updates @swathi-sadananda has put into the docs if you haven't already to ensure that we properly document how someone can do an incremental build.
Once that's in place, then lets see if anything is required here.

jbyibm:
Not needed. If we have issues, then we must open a bug for the DBB part. For @davidjwriter

PR #124, opened 2026-06-29
Deploy frontend to standalone liberty instance

AndrewTwydell:
This migrates the frontend application deployment from the z/OS connect server to it's own standard liberty instance. It's all configured in the config.yaml, like everything else, and lifecycles the same as the zosconnect instance.

API is still on `:9080/api` by default.
Frontend is `:9081` by default.

`http://<host>:9081/admin.html` queries `http://<host>:9080/api/customers/1234`

### DIFF
--- a/.setup/config/config.yaml
+++ b/.setup/config/config.yaml
@@ -89,6 +89,14 @@ zosconnect:
   http_port: 9080
   https_port: 9443
 
+# Frontend Liberty Server (separate from z/OS Connect)
+frontend:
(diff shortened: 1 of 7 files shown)


jbyibm:
@davidjwriter   can you review I'll not have time.

davidjwriter:
@jbyibm yes I'm working on it this afternoon

davidjwriter (approved):
I deployed it and seems like everything works! Frontend is reachable. I ran into errors trying to create a new customer, but I think that error was there before.

PR #125, opened 2026-06-30
Delete Customer API

### LINKED ISSUE #108: Delete Customer API
API to delete customers needs to be implemented.

Sash9:
- Implements the Delete Customer API, enabling customers to be permanently removed from the system via a DELETE /customers/{customerId} request. 
- The Get Customer endpoint has also been updated to correctly return a 404 when a customer does not exist
- Delete Customer page in the frontend has been updated to handle the standard customer ID format used across the application.

Screenshot 2026-06-30 at 12 50 33 AM


MikeFultonDev:
@AndrewTwydell please review asap - Sashank will be finishing up soon and I'd like to see this completed

AndrewTwydell (approved):
The customer is deleting successfully on the backend, but I'm getting an error in the GUI saying it failed. Yet when re-searching for the customer, they don't exist. 

Going to investigate to see if it's a bad HTTP code coming back from z/OS connect - perhaps the 204 condition is not being met - or if the frontend is not handling a 204 as expected.

I would be happy for this to go in and we can follow up with additional error checking.

PR #139, opened 2026-06-30
Debug cics

akarputkin:
The change primarily focuses on enabling CICS TCP/IP sockets interface for CICSBOZ region - a prerequisite for z/OS debugger. There are 2 jcl-s that will create necessary datasets, a change in the setup script to run these jcl-s, extra parameters for the cics region to enable CICS socket interface and include one of the output datasets (DFHTABLE DD), and a lot of extra definition in the SOCKETS group in debug-definitions.yaml.

Also there's a change for cics.xml in zosconnect, increasing the request timeout for CICS requests to 5 mins as the default value of 30s makes debugging unusable.


jbyibm (approved):
Looks Good

PR #150, opened 2026-07-07
Check EQARMTD; add a debugging tutorial

akarputkin:
The fix checks if EQARMTD is already running before starting the task and adds a debugging tutorial


docs/docs/tutorials/debug-cics-transaction.md
```diff
@@ -0,0 +1,167 @@
+---
+layout: default
+title: Debug a CICS Transaction
+---
+# Debug a CICS Transaction
+
+## Overview
+
+This tutorial shows how to debug the Bank of Z `INQCUST` COBOL program — the CICS customer inquiry transaction — using IBM Z Open Debug in Visual Studio Code together with the **Debug Profile Service (DPS)**. The Debug Profile Service routes a debug session to your specific VS Code instance without affecting other users of the same CICS region.
```
AndrewTwydell: I was under the impression that's the role of the Remote debug service. 

It’s the Remote Debug Service that routes the live debug connection, while DPS manages the profiles that tell it what to do. 

Unless I've misunderstood what you're trying to say here.

docs/docs/tutorials/debug-cics-transaction.md
```diff
@@ -0,0 +1,167 @@
...
+```
+VS Code  ⟷  Remote Debug Server (RDS, port 8194)  ⟷  Debug Profile Service (DPS, port 8192)  ⟷  CICS region
+```
+
+The **Debug Profile Service** holds named profiles. When a CICS transaction runs under a user ID that matches a profile, CICS suspends the task and notifies the RDS, which connects it to the VS Code debug session. Only transactions belonging to TSO user ID specified in the profile are intercepted — other users are unaffected.
+
+| Component | Role | Port (Bank of Z default) |
+|-----------|------|--------------------------|
+| Debug Profile Service (DPS) | Stores and serves debug profiles | 8192 |
+| Remote Debug Server (RDS) | Protocol bridge between DPS and VS Code DAP | 8194 |
```
AndrewTwydell: What is the DAP acronym? Is that defined somewhere earlier in the document?

docs/docs/tutorials/debug-cics-transaction.md
```diff
@@ -0,0 +1,167 @@
...
+
+Select **Save**. The profile is registered with the DPS on z/OS.
+
+**Step 3.4:** Activate the profile by right-clicking it and selecting **Activate for Debug**. The icon of a bug with a green dot appears at the profile when active.
+
+With the profile active, any invocation of `INQCUST` under `CICSUSER` (the ID used by zOS Connect) user ID is intercepted by CICS and parked until VS Code attaches.
+
+## Step 4 — Trigger the CICS Transaction
+
+Invoke the `OMEN` transaction so that CICS runs `INQCUST` under your user ID.
```
AndrewTwydell: I believe the idea is to remove the BMS map interface eventually. Should we tailor the tutorials towards the web interface to start with, or are we planning to come back and change them after?

AndrewTwydell: Ah I see underneath you're providing both options. 

A user doesn't run the OMEN transaction when using the web UI. zOS connect runs the programs directly.

docs/docs/tutorials/debug-cics-transaction.md
akarputkin: you're right, I will rephrase this.

docs/docs/tutorials/debug-cics-transaction.md
akarputkin: DAP is Debug Adapter Protocol, that's how VS Code talks to debuggers. RDS is our DAP server, it translates our internal protocol into DAP

docs/docs/tutorials/debug-cics-transaction.md
akarputkin: should I remove the 3270 option?

MikeFultonDev: Lets make this tutorial just for debug of the Web UI and then we can consider adding a 3270 debug session in the future if people want to see it - it'll keep this more streamlined

akarputkin: done, I removed 3270 and added a small note on timeouts for the web interface

MikeFultonDev:
LGTM but I'd like @AndrewTwydell to approve

AndrewTwydell (approved):

PR #161, opened 2026-07-13
frontend liberty server setup

adn1107:
fix: resolve FRONTEND_LIBERTY_HOME in frontend Liberty server setup

## Problem
Stage 11 fails with:
  setup-frontend-server.sh: line 54: /bin/server: EDC5129I No such file or directory

`FRONTEND_LIBERTY_HOME` is empty at runtime. Commit 7c1d5a9 renamed
`LIBERTY_HOME` → `FRONTEND_LIBERTY_HOME` and moved the export into
`setenv.sh`, but left 3 call-sites in `setup-frontend-server.sh` still
using the old name. Separately, the `.env` cache is only invalidated
when `config.yaml` changes — not when `setenv.sh` itself changes — so
systems with a cached `.env` from before 7c1d5a9 never pick up the new
variable.

## Changes
- `setup-frontend-server.sh`: replace 3 stale `LIBERTY_HOME` refs with `FRONTEND_LIBERTY_HOME`
- `setenv.sh`: add `|| "$ENV_FILE" -ot "${BASH_SOURCE[0]}"` to the staleness check so `.env` is regenerated whenever `setenv.sh` changes

## Testing
- Fresh env: stage 11 completes, server created at expected path
- Re-run: server stopped before `rm -rf`, no CWWKE0045E
- Stale `.env`: regenerated on first run, `FRONTEND_LIBERTY_HOME` populated correctly

Signed-off-by: Andrew Nguyen <andy.nguyen@ibm.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ scripts/update-frontend-file.sh
 scripts/update-frontend-file-zos.sh
 **/.taz-edt/
 **/datasets.yaml
+.gradle/
--- a/.setup/config/setenv.sh
+++ b/.setup/config/setenv.sh
@@ -35,7 +35,7 @@ fi
 set -e
 
 
-if [[ ! -f "$ENV_FILE" || "$ENV_FILE" -ot "$CONFIG_FILE" ]]; then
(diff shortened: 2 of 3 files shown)


AndrewTwydell (approved):
Nice one, I hit this too

PR #162, opened 2026-07-13
Use product name for zconfig and add to glossary

andrewhughes101:
Uses the correct capitalization for zconfig
Adds the full name for zconfig to the pre-reqs
Adds a glossary entry for zconfig using the description from the zconfig docs


AndrewTwydell (approved):
Nice one

PR #165, opened 2026-07-15
Feature: HTTPS Support for Bank of Z

QihanQiu1:
Configures HTTPS for the Bank of Z web application using a RACF keyring and a re-signed server certificate that satisfies modern browser requirements (EKU serverAuth, SANs, CAB Forum validity limits).

What this does:

- RACF keyring ${ZOS_KEYRING} holds the server certificate (BoZ, signed by ${ZOS_CA_LABEL}). All values come from config.yaml via setenv.sh — nothing hardcoded.
- gencert-eku.sh uses a re-sign approach: RACF generates the keypair and initial cert; Bouncy Castle re-signs it adding EKU TLS Web Server Authentication and SANs. The private key never leaves RACF.
- Cert validity is capped at the CAB Forum SC-081 step-down schedule (200 days until 2027-03-14, 100 days until 2029-03-14, 47 days after)
- Liberty uses safkeyring://${ZOS_ADMIN_USER}/${ZOS_KEYRING} (JCERACFKS) — no keystore files on disk.
- Two-server design: BAQ${APP_SHORT_NAME} (port 9444 HTTPS / 9080 HTTP) serves the API WAR at /api; FE${APP_SHORT_NAME} (port 9445 HTTPS / 9081 HTTP) serves the frontend WAR at /. config.js selects the API base URL based on the page protocol
- All hardcoded values (IBMUSER, BOZRING, VSICA, ASM HLQ, IMS macro lib) replaced with variables from config.yaml.


.setup/config/config.yaml
```diff
@@ -130,6 +132,7 @@ ims:
   db2_plan_ibshist: "IBSHIST"
   jobcard: "//{{jobname if jobname is defined else 'BOZJOB'}} JOB (ACCT),'IMS BANK',CLASS=A,MSGCLASS=H,NOTIFY=&SYSUID"
   asm_lib_hlq: "ASM.SASMMOD1"
+  ims_macro_lib: "IMSV15.SDFSMAC"
```
ddimatos: This feels like it should not be hardcoded, there is going to be an IMS16 , just a thought.

ddimatos: I realize it is a config file, while the last qualifier is constant `SDFSMAC` the HLQ will change based on relationship to IMS version `IMSV15`or default `IMS` as the HLQ.

.setup/setup-common.sh
```diff
@@ -330,6 +330,30 @@ stage_populate_ims_database() {
...
+        print_error "Certificate script not found: $BANK_DIR/.setup/setup/addcert.sh"
+        exit 1
+    fi
+
+    print_info "Running Bank of Z certificate setup script..."
+    print_info "Executing: bash $BANK_DIR/.setup/setup/addcert.sh"
+    cd "$BANK_DIR"
+
+    set -o pipefail
+    if bash .setup/setup/addcert.sh; then
```
ddimatos, line 353: Are we absolutely dependent on Bash vs z/OS default shell? I see the shebang `#!/usr/bin/env bash` which is fine, but maybe something to consider in the future is posix compliance.

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# The actual server cert (with EKU serverAuth for Safari compliance) is
+# generated by gencert-eku.sh using a CSR round-trip:
+#   - RACDCERT GENCERT REQONLY  → keypair + CSR (private key stays in RACF)
+#   - Bouncy Castle signs the CSR, adds EKU + SANs
+#   - RACDCERT IMPORT returns the signed cert to the keyring
+# =============================================================================
+
+## CUSTOMIZE ##
+userid=IBMUSER
```
ddimatos: @MikeFultonDev is ok to have user names out in the open like this, future considering if scripting you can encrypt it in a local file, or us a vault (hashicorp vault) or with ansible ansible vault or AAP credential manager.

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+# Step 2: Generate the server cert with EKU serverAuth via Bouncy Castle CSR
+# round-trip (gencert-eku.sh).  The private key is generated in RACF and never
+# leaves the keyring.
+if test $rc -eq 0
+then
+  SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+  if sh "$SCRIPT_DIR/gencert-eku.sh"; then
+    echo "EKU cert generation succeeded — cert in keyring IBMUSER/$ring"
+  else
+    echo "WARNING: gencert-eku.sh failed — Liberty may start without a server cert" >&2
```
ddimatos: Some logic feels off here, line [352](https://github.com/IBM/Bank-of-Z/pull/165/files#diff-ee029e2a11478fe867924406448d9a5c2c2cf8b6545e09fb364fefd63151135eR350) print_success "RACF keyring and certificates created successfully" in setup-common.sh  says it was successful in setup of RACF keyring and certs giving false understand that at this juncture, you could actually still fail to start Liberty without a cert. Feels like you should have bubbled up the error.

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
-frontend.https.port=${FRONTEND_HTTPS_PORT}
-zosconnect.http.port=${ZOSCONNECT_HTTP_PORT}
-EOF
+BOOTSTRAP_DEST="${WLP_USER_DIR}/servers/${SERVER_NAME}/bootstrap.properties"
+$PYTHON -c "
+content = '# Frontend Liberty Server Bootstrap Properties\nfrontend.http.port=${FRONTEND_HTTP_PORT}\nfrontend.https.port=${FRONTEND_HTTPS_PORT}\nzosconnect.http.port=${ZOSCONNECT_HTTP_PORT}\n'
+with open('${BOOTSTRAP_DEST}', 'wb') as f:
+    f.write(content.encode('iso-8859-1'))
+"
+chtag -t -c ISO8859-1 "${BOOTSTRAP_DEST}"
```
ddimatos: FYI, you could have avoided the python and just run `iconv -f IBM-1047 -t ISO8859-1 "$file">"$new_file"`.

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
 
-# Deploy API WAR file configuration
-# ==================# =========================
-# Configure CORS for frontend server
 # =========================
-cat > "${WLP_USER_DIR}/servers/${APP_BASE_NAME_LOWER}Server/configDropins/overrides/cors.xml" << EOF
-<?xml version="1.0" encoding="UTF-8"?>
-<server description="CORS configuration for frontend server">
+# Deploy SSL/TLS configuration (RACF keyring with EKU cert).
+# Written via Python to avoid _BPXK_AUTOCVT EBCDIC corruption on cp.
```
ddimatos: Its not really corruption if you know what his happening, shell utils will default the shells encoding, in this case its EBCDIC, so it is not corrupted its what was cp was asked to do. You need to follow up with a util like `icov` to get it to the target ASCII encoding. Given public exposure, i'd suggest another word over corruption.

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
+              location=\"safkeyring://IBMUSER/BOZRING\"
+              type=\"JCERACFKS\"
+              password=\"password\"/>
+</server>
+'''
+with open('${TLS_DEST}', 'wb') as f:
+    f.write(content.encode('iso-8859-1'))
+"
+chtag -t -c ISO8859-1 "${TLS_DEST}"
+print_success "SSL configuration deployed (safkeyring://IBMUSER/BOZRING)"
```
ddimatos: How do you know success has been reached, a Python process is started and no validation afterwards.

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
+HTTPS_PORT=$(get_section_value 'zosconnect' 'https_port')
+HTTPS_PORT=${HTTPS_PORT:-9444}
+HTTP_EP_DEST="$OVERRIDES_DIR/http-endpoint.xml"
+$PYTHON -c "
+content = '<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<server>\n    <httpEndpoint id=\"defaultHttpEndpoint\" host=\"*\" httpPort=\"9080\" httpsPort=\"${HTTPS_PORT}\"/>\n</server>\n'
+with open('${HTTP_EP_DEST}', 'wb') as f:
+    f.write(content.encode('iso-8859-1'))
+"
+chtag -t -c ISO8859-1 "$HTTP_EP_DEST"
+print_success "HTTPS port override deployed (httpsPort=${HTTPS_PORT})"
```
ddimatos: Pattern repeats, I will mention once more but not everywhere, I see many places in the code where success is assumed not verified.

src/frontend/server.js
```diff
@@ -105,11 +108,28 @@ function proxyApiRequest(req, res) {
...
+const SSL_CERT = process.env.SSL_CERT; // path to PEM cert file
+const SSL_KEY  = process.env.SSL_KEY;  // path to PEM key file
+
+if (SSL_CERT && SSL_KEY) {
+    const tlsOptions = {
+        cert: fs.readFileSync(SSL_CERT),
+        key:  fs.readFileSync(SSL_KEY)
+    };
+    server = https.createServer(tlsOptions, requestHandler);
+    console.log(`Server will run at https://0.0.0.0:${PORT}/`);
```
ddimatos: Minor thing, I usually see localhost/0.0.0.0 as to where a system is listening on regarding a port, server will run on the value returned by `hostname`:${PORT}, its minor.

ddimatos (changes requested):
Requesting changes for the reason that there appears to be quite a few operations taken without validation, in some cases success is assumed and not validated. 

There are some additional comments around the usage of python when it could have remained as shell, those are just comments as I am unsure how familiar with z/OS the developer is. 

Some other points about a self signed certificate I will take offline, the cert and signing process seems correct, nothing stuck out at me although its been a while. 

Thank you, this is good work.

.setup/config/config.yaml
MikeFultonDev: @jbyibm please provide input here. Probably 'ASM.' needs the same update suggested for IMS

.setup/setup/addcert.sh
MikeFultonDev: should use the right env var - see also up in tls.xml - @jbyibm what is the right env var to reference here?

.setup/setup/addcert.sh
MikeFultonDev: need to check return codes here. I think something called 'addcert' probably shouldn't be deleting a cert as a side effect - perhaps this should be broken into 2 scripts? 
At the start, should probably check that the user running has the authority to make these changes - and the user running should be the userid used - but I'd like @jbyibm 's opinion and @ddimatos 's on this.

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
+              type=\"JCERACFKS\"
+              password=\"password\"/>
 
 </server>
-EOF
+'''
+with open('${SERVER_XML_DEST}', 'wb') as f:
+    f.write(content.encode('iso-8859-1'))
+"
+chtag -t -c ISO8859-1 "${SERVER_XML_DEST}"
```
MikeFultonDev: is the tag required here? I would have expected Python to tag the file as ISO8859-1 since you did a content.encode. If it doesn't, add a comment to indicate that is why the tag is required.

.setup/setup/setup-frontend-server.sh
MikeFultonDev: if you use iconv, be explicit and say /bin/iconv to ensure you get the UNIX System Services built-in iconv. There is also an option on /bin/iconv that will perform the tagging for you so you don't have to issue a separate chtag. 
I would encourage you to use 'bash' for this script rather than 'sh' and if you do that, cat and echo will automatically create the file in ISO8859-1 (and tag it for you) and you can avoid this hassle.

.setup/setup/setup-zosconnect-server.sh
MikeFultonDev: I think using bash can simplify this. If it still causes you grief, I'm ok with you doing it all in Python alternately. Either of those is going to be simpler (and faster) than bouncing between 'sh' and Python. 
If you use bash, do it in a consistent way with how others have done it - the simplest is just to add 
`#!/bin/env bash`
on the first line of the script but some teams prefer a more explicit:

bash <script> 

to invoke the script.

MikeFultonDev (commented):
I have some comments.
This is a great first pass @QihanQiu1 ! This is a complex problem.

.setup/config/config.yaml
```diff
@@ -73,21 +73,23 @@ wazideploy:
...
-# z/OS Connect (API exposure)
+# z/OS Connect (API exposure + frontend static files)
+# Single Liberty instance hosts both api.war (/api) and bank-frontend-vanilla.war (/)
 zosconnect:
   zosconnect_home: "/usr/lpp/IBM/zosconnect/bin"
   server_dir: "{{sandbox.path}}/zosconnect-server"
   server_name: "bankOfZServer"
   http_port: 9080
-  https_port: 9443
+  https_port: 9444
```
OnnoVdT: why change port number?

.setup/config/config.yaml
```diff
@@ -73,21 +73,23 @@ wazideploy:
...
-# Frontend Liberty Server (separate from z/OS Connect)
+# Frontend Liberty Server — no longer used; frontend is hosted on z/OS Connect
+# https_port is 9445 (not 9444) to avoid conflict with BAQBANKZ on 9444
 frontend:
   liberty_home: "/usr/lpp/liberty_zos/25.0.0.9"
   server_dir: "{{sandbox.path}}/frontend"
   server_name: "bankz-frontend"
   http_port: 9081
-  https_port: 9444
+  https_port: 9445
```
OnnoVdT: why change port number?

.setup/config/config.yaml
OnnoVdT: All HLQs must be variables. And I would say the same is true for paths, but here things do tend to stick to the default more often so it's less critical.

.setup/jcl/ims/Ims-dynalloc.j2
```diff
@@ -116,7 +116,7 @@
 //IMSDALOC  PROC  SOUT=*
 //ASSEM  EXEC PGM=ASMA90,PARM='ALIGN,DECK,NOOBJECT'
 //STEPLIB  DD DSN={{ ims.asm_lib_hlq }},DISP=SHR
-//SYSLIB   DD DISP=SHR,DSN={{ ims.ims_hlq }}.SDFSMAC
+//SYSLIB   DD DISP=SHR,DSN={{ ims.ims_macro_lib }}
```
OnnoVdT: Be consistent with variable usage, earlier you were moving from hardcoded DSN to {{ ims.ims_sys_hlq }}.SDFSMAC.
Personally, I think that just the HLQ as variable with hardcoded LLQ is ideal for SMPE managed libraries (since these have fixed LLQs) as that reduces the number of variables, and thus makes the config file less intimidating. For libraries created during customization this rule is more of a guideline as here LLQ is also variable (but we could drive users to use fixed LLQs so we can use a limited number of HLQ vars and keep the customization work minimal and crisp).

.setup/setup/zosconnect-ssl/http-endpoint.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<server>
+    <httpEndpoint id="defaultHttpEndpoint" host="*" httpPort="9080" httpsPort="9444"/>
```
OnnoVdT: Shouldn't we be pulling the ports from config.yaml?

.setup/setup/zosconnect-ssl/tls.xml
```diff
@@ -0,0 +1,18 @@
...
+      round-trip: keypair generated in RACF, signed by VSICA via Bouncy Castle
+      with EKU serverAuth added, then imported back. Private key never leaves RACF.
+    -->
+    <featureManager>
+        <feature>ssl-1.0</feature>
+        <feature>transportSecurity-1.0</feature>
+    </featureManager>
+    <ssl id="defaultSSLConfig" keyStoreRef="defaultKeyStore"/>
+    <keyStore id="defaultKeyStore"
+              location="safkeyring://IBMUSER/BOZRING"
```
OnnoVdT: in addcert.sh, IBMUSER and BOZRING are in the "customize" section, so shouldn't the be veribles both here and there?

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+
+## CUSTOMIZE ##
+userid=IBMUSER
+ring=BOZRING
+
+## FIXED ##
+profile=$userid.$ring.LST
+
+# safety net, not sure if class is active or not
+tsocmd "SETROPTS GENERIC(SERVAUTH)" \
```
OnnoVdT: why SERVAUTH? Shouldn't this be RDATALIB?

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+ >/dev/null 2>&1
+tsocmd "SETROPTS CLASSACT(SERVAUTH) RACLIST(SERVAUTH)" \
```
OnnoVdT: same as before

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+ >/dev/null 2>&1
+
+# cleanup, might fail
+tsocmd "RACDCERT ID($userid) DELRING($ring)" \
```
OnnoVdT: we should add a note saying gencert-eku.sh will delete cert itself

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+tsocmd "SETROPTS RACLIST(DIGTCERT DIGTRING) REFRESH"
+
+# Step 2: Generate the server cert with EKU serverAuth via Bouncy Castle CSR
+# round-trip (gencert-eku.sh).  The private key is generated in RACF and never
+# leaves the keyring.
+if test $rc -eq 0
+then
+  SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+  if sh "$SCRIPT_DIR/gencert-eku.sh"; then
+    echo "EKU cert generation succeeded — cert in keyring IBMUSER/$ring"
```
OnnoVdT: $userid instead of IBMUSER

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+#   1. RACDCERT GENCERT SIGNWITH(CERTAUTH LABEL('VSICA')) — RACF generates
+#      keypair and a VSICA-signed cert.  Private key stays in RACF.
+#   2. dcp export the cert as CERTB64 (PEM) to USS.
+#   3. Bouncy Castle reads the public key from the PEM, builds a new
+#      TBSCertificate with EKU serverAuth + SANs + 397-day validity, and
+#      re-signs it with VSICA's key (exported as PKCS12DER, deleted after).
+#   4. dcp the new DER cert back to a RACF dataset.
+#   5. RACDCERT ADD FORMAT(CERTDER) — RACF matches the new cert to the
+#      private key it already holds (same public key).
+#   6. Connect cert to BOZRING as DEFAULT.
```
OnnoVdT: I would move add to keyring to caller (addcert.sh) as it is not a gencert action

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+# (setenv.sh requires bash; this script is sh — do not source it here.)
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+if [ -z "${SANDBOX_DIR:-}" ]; then
+  SANDBOX_DIR="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+  export SANDBOX_DIR
+fi
+
+set -e
+
+userid=IBMUSER
```
OnnoVdT: should be a passed-in variable

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+ring=BOZRING
+label='BoZ'
+
+# -----------------------------------------------------------------------
+# Resolve tools — fall back to known fixed paths when env vars not set.
+# -----------------------------------------------------------------------
+if [ -z "${JAVA_HOME:-}" ]; then
+  JAVA_HOME=/usr/local/sandboxes/tools/J21.0_64
```
OnnoVdT: shouldn't this be a variable as well? 
[https://ibm-systems-z.slack.com/archives/C0ALLEG5SP3/p1780296791820699?thread_ts=1779986154.891589&cid=C0ALLEG5SP3](https://ibm-systems-z.slack.com/archives/C0ALLEG5SP3/p1780296791820699?thread_ts=1779986154.891589&cid=C0ALLEG5SP3)
* The sandbox folder is not hardcoded at all in our current scripts
* The tools scripts create the new volume for the sandbox if not there

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+  PYTHON=/usr/lpp/IBM/cyp/v3r14/pyz/bin/python3
+else
+  PYTHON=$(command -v python3 2>/dev/null) || { echo "[gencert-eku] FATAL: python3 not found" >&2; exit 1; }
+fi
+
+# Ensure ZOAU tools (dcp, tsocmd, etc.) are on PATH
+if [ -n "${ZOAU_HOME:-}" ] && [ -x "$ZOAU_HOME/bin/dcp" ]; then
+  export PATH="$ZOAU_HOME/bin:$PATH"
+elif [ -x /usr/lpp/IBM/zoautil/bin/dcp ]; then
+  export PATH="/usr/lpp/IBM/zoautil/bin:$PATH"
```
OnnoVdT, line 76: ZOAU 1.4 (and it looks like 1.3 as well) switched to default path /usr/lpp/IBM/zoau

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+  | awk '/End Date:/ {gsub("/","-",$3); print $3}')
+
+echo "[gencert-eku] IP=$ipaddr  DNS=$dnsname  VSICA expire=$expire"
+echo "[gencert-eku] Private key stays in RACF keyring throughout."
+
+# Random passwords via Python
+CA_PASS=$($PYTHON -c "import secrets; print(secrets.token_urlsafe(18))")
+
+# Race-safe temp dir
+TMPDIR=/tmp/boz-cert-$$
```
OnnoVdT: do not hardcode /tmp, use ${TMPDIR:-/tmp} so a non-default /tmp can be specified before calling the script. I know some customers do not allow users to write to /tmp, only servers can, users have a separate temp filesys.

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+mkdir -m 700 -p "$TMPDIR"
```
OnnoVdT: mkdir -m is still going through umask, so you might not get 700. Either first set umask or use chmod after mkdir to ensure you end up with 700.

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+trap 'rm -rf "$TMPDIR"
+  tsocmd "DELETE (\047${userid}.BOZ.CAKEY\047)" >/dev/null 2>&1 || true
```
OnnoVdT, line 113: side note, you can also use `mv "//'dsn'" /dev/null` to delete a dataset

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+  tsocmd "DELETE (\047${userid}.BOZ.CERTB64\047)" >/dev/null 2>&1 || true
+  tsocmd "DELETE (\047${userid}.BOZ.NEWCERT\047)" >/dev/null 2>&1 || true' EXIT
+
+# -----------------------------------------------------------------------
+# 1. Remove any existing cert so GENCERT can create a fresh one.
+# -----------------------------------------------------------------------
+echo "[gencert-eku] Removing old cert from keyring..."
+tsocmd "RACDCERT ID($userid) \
```
OnnoVdT: ring is already deleted by caller, so this will always fail

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+ >/dev/null 2>&1
+tsocmd "RDELETE RDATALIB $profile" \
+ >/dev/null 2>&1
+tsocmd "SETROPTS RACLIST(RDATALIB) REFRESH" \
+ >/dev/null 2>&1
+
+# Step 1: Create the keyring and connect VSICA as the trust anchor.
+tsocmd "RACDCERT ID($userid) ADDRING($ring)"
+tsocmd "RACDCERT ID($userid) \
+ CONNECT(CERTAUTH LABEL('VSICA') RING($ring) USAGE(CERTAUTH))"
```
OnnoVdT: make VSICA a variable, this is only valid on ZVDT systems.

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+test "$ipaddr" = "IntfName:" && ipaddr=$(netstat -h 2>/dev/null \
+  | awk '/ OSA/ {f=1; next} f {print $2; exit}')
+dnsname=$(hostname 2>/dev/null)
+
+if [ -z "$ipaddr" ] || [ -z "$dnsname" ]; then
+  echo "[gencert-eku] FATAL: could not determine IP ($ipaddr) or DNS ($dnsname)" >&2
+  exit 1
+fi
+
+expire=$(tsocmd "RACDCERT CERTAUTH LIST(LABEL('VSICA'))" \
```
OnnoVdT: make VSICA a variable, this is only valid on ZVDT systems.

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+# safety net, not sure if class is active or not
+tsocmd "SETROPTS GENERIC(SERVAUTH)" \
+ >/dev/null 2>&1
+tsocmd "SETROPTS CLASSACT(SERVAUTH) RACLIST(SERVAUTH)" \
+ >/dev/null 2>&1
+
+# cleanup, might fail
+tsocmd "RACDCERT ID($userid) DELRING($ring)" \
+ >/dev/null 2>&1
+tsocmd "SETROPTS RACLIST(DIGTCERT DIGTRING) REFRESH" \
```
OnnoVdT: DIGTCERT has not changed, so refresh just burns resources

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,73 @@
...
+ >/dev/null 2>&1
+tsocmd "SETROPTS RACLIST(RDATALIB) REFRESH" \
+ >/dev/null 2>&1
+
+# Step 1: Create the keyring and connect VSICA as the trust anchor.
+tsocmd "RACDCERT ID($userid) ADDRING($ring)"
+tsocmd "RACDCERT ID($userid) \
+ CONNECT(CERTAUTH LABEL('VSICA') RING($ring) USAGE(CERTAUTH))"
+rc=$?
+tsocmd "SETROPTS RACLIST(DIGTCERT DIGTRING) REFRESH"
```
OnnoVdT: DIGTCERT has not changed, so refresh just burns resources

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+import java.time.ZoneOffset;
+import java.util.Base64;
+import java.util.Date;
+
+public class ResignCert {
+    static final ASN1ObjectIdentifier EKU_SERVER_AUTH =
+        new ASN1ObjectIdentifier("1.3.6.1.5.5.7.3.1");
+    static final ASN1ObjectIdentifier SHA256_WITH_RSA =
+        new ASN1ObjectIdentifier("1.2.840.113549.1.1.11");
+    static final long MAX_VALIDITY_MS = 397L * 24 * 60 * 60 * 1000;
```
OnnoVdT: Leave a comment that this is to set 397 day validity for the cert (which by the way is already out of bound for international standards). With the validity window drastically shrinking in the next 2 years it must be easy to spot what to change to get a lower limit. Or perhaps already define the proposed limits as constants and pick the right constant based on the current date so no more code changes are needed.

Publicly trusted SSL/TLS certificates have a maximum validity period of 199–200 days (roughly 6 months). This lifespan will step down to 100 days in March 2027 and reach 47 days in March 2029. (Mandated by the CA/Browser Forum.)

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+
+# Derive SANDBOX_DIR from script location when not already in environment.
+# (setenv.sh requires bash; this script is sh — do not source it here.)
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+if [ -z "${SANDBOX_DIR:-}" ]; then
+  SANDBOX_DIR="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+  export SANDBOX_DIR
+fi
+
+set -e
```
OnnoVdT: add comment that with this we will exit on non=zero rc as explicit reminder for those looking at the code

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+
+# -----------------------------------------------------------------------
+# 7. Copy new DER cert to RACF dataset and IMPORT it.
+#    RACF will match it to the private key already held for label '$label'
+#    because they share the same public key.
+# -----------------------------------------------------------------------
+echo "[gencert-eku] Importing re-signed cert into RACF..."
+# Pre-allocate with correct DCB for binary DER data
+tsocmd "DELETE ('${userid}.BOZ.NEWCERT')" >/dev/null 2>&1 || true
+tsocmd "ALLOC DATASET('${userid}.BOZ.NEWCERT') NEW CATALOG \
```
OnnoVdT: stick with explicit DELETE and ALLOC NEW if dcp appends, but if it overwrites I would skip DELETE and use ALLOC MOD

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+# dcp for binary copy (cp "//dataset" doesn't work reliably for DER)
+$DCP -B "$TMPDIR/boz-new.der" "${userid}.BOZ.NEWCERT"
+
+tsocmd "RACDCERT ADD('${userid}.BOZ.NEWCERT') \
+  ID($userid) \
+  WITHLABEL('$label') \
+  FORMAT(CERTDER) \
+  TRUST"
+
+tsocmd "RACDCERT ID($userid) \
```
OnnoVdT: move to caller, does not belong in gencert

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,351 @@
...
+
+tsocmd "SETROPTS RACLIST(DIGTCERT DIGTRING) REFRESH"
+
+echo "[gencert-eku] Done."
+echo "[gencert-eku]   Cert label : $label  (in keyring IBMUSER/$ring)"
+echo "[gencert-eku]   Private key: stays in RACF — never written to USS filesystem"
+echo "[gencert-eku]   Validity   : 397 days (Safari/Apple ATS compliant)"
+echo "[gencert-eku]   SANs       : IP=$ipaddr  DNS=$dnsname"
+echo "[gencert-eku]   EKU        : TLS Web Server Authentication"
+echo "[gencert-eku]   Liberty    : uses safkeyring://IBMUSER/$ring (JCERACFKS)"
```
OnnoVdT: $userid instead of IBMUSER

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
         <feature>transportSecurity-1.0</feature>
+        <feature>ssl-1.0</feature>
     </featureManager>
 
     <!-- HTTP Endpoint Configuration -->
-    <httpEndpoint id="defaultHttpEndpoint"
-                  httpPort="${frontend.http.port}"
-                  httpsPort="${frontend.https.port}"
-                  host="*" />
+    <httpEndpoint id=\"defaultHttpEndpoint\"
```
OnnoVdT: additional defaultHttpEndPoint options to consider
`    <!--onError="FAIL"-->`
`    <!--portOpenRetries="10"--> <!-- for startup at IPL -->`
[ https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-httpendpoint]( https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-httpendpoint)

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
-             maxFileSize="20" 
-             maxFiles="10" />
+    <logging traceSpecification=\"*=info\"
+             maxFileSize=\"20\"
+             maxFiles=\"10\" />
 
-    <!-- SSL Configuration (optional - for HTTPS) -->
-    <keyStore id="defaultKeyStore" password="Liberty" /> <!-- pragma: allowlist secret -->
+    <!-- SSL Configuration using RACF keyring -->
+    <ssl id=\"defaultSSLConfig\" keyStoreRef=\"defaultKeyStore\"/>
```
OnnoVdT: additional ssl options to consider
`    sslProtocol=""`
`    enabledCiphers=""`
[https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-ssl](https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-ssl)

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
+    <keyStore id=\"defaultKeyStore\"
+              location=\"safkeyring://IBMUSER/BOZRING\"
+              type=\"JCERACFKS\"
+              password=\"password\"/>
 
 </server>
-EOF
+'''
```
OnnoVdT: more definitions to consider
```
  <!-- Hide Liberty welcome page
    https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-httpdispatcher -->
  <httpDispatcher
    enableWelcomePage="false"
    />

  <!-- Avoid idle CPU usage
    https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-config -->
  <config
    updateTrigger="disabled"
    />

  <!-- Avoid idle CPU usage
    https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-applicationmonitor -->
  <applicationMonitor
    dropinsEnabled="false"
    updateTrigger="disabled"
    />

  <!-- Server configuration security hardening
    https://www.openliberty.io/docs/22.0.0.1/server-configuration-hardening.html
    https://www.ibm.com/docs/en/was-liberty/core?topic=configuration-webcontainer -->
  <webContainer
    disableXPoweredBy="true"
    />

  <!-- Avoid 404 page due to slow startup
    https://www.ibm.com/docs/en/was-liberty/nd?topic=configuration-applicationmanager -->
  <!--applicationManager
    startTimeout="30s"
    /-->

```

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
+$PYTHON -c "
+import sys
+content = '''<?xml version=\"1.0\" encoding=\"UTF-8\"?>
+<server description=\"IPIC connection to CICS\">
     <featureManager>
         <feature>zosconnect:cics-1.0</feature>
     </featureManager>
-    <zosconnect_cicsIpicConnection id="${APP_BASE_NAME_LOWER}CicsConnection" host="127.0.0.1" port="${CICS_IPIC_PORT}" sysid="ZC01" authDataRef="cicsCredentials" requestTimeout="300s" />
-    <zosconnect_authData id="cicsCredentials" user="${CICS_USER}" password="${CICS_PASSWORD}" />
+    <zosconnect_cicsIpicConnection id=\"${APP_BASE_NAME_LOWER}CicsConnection\" host=\"127.0.0.1\" port=\"${CICS_IPIC_PORT}\" sysid=\"ZC01\" authDataRef=\"cicsCredentials\" requestTimeout=\"300s\" />
```
OnnoVdT: should we make 127.0.0.1 a variable so it's easy to switch when used in a IPv6-only environment?

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
+      imported back. Private key never leaves RACF.
+    -->
     <featureManager>
-        <feature>cors-1.0</feature>
+        <feature>ssl-1.0</feature>
+        <feature>transportSecurity-1.0</feature>
     </featureManager>
+    <ssl id=\"defaultSSLConfig\" keyStoreRef=\"defaultKeyStore\"/>
+    <keyStore id=\"defaultKeyStore\"
+              location=\"safkeyring://IBMUSER/BOZRING\"
```
OnnoVdT: use variables for userid & keyring

.setup/setup/setup-zosconnect-server.sh
```diff
@@ -104,58 +104,130 @@ rm -f "/tmp/BAQ${APP_BASE_NAME}.jcl"
...
+              type=\"JCERACFKS\"
+              password=\"password\"/>
```
OnnoVdT: add comment saying keyring password must be "password", but is not actually used.

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
+             maxFiles=\"10\" />
 
-    <!-- SSL Configuration (optional - for HTTPS) -->
-    <keyStore id="defaultKeyStore" password="Liberty" /> <!-- pragma: allowlist secret -->
+    <!-- SSL Configuration using RACF keyring -->
+    <ssl id=\"defaultSSLConfig\" keyStoreRef=\"defaultKeyStore\"/>
+    <keyStore id=\"defaultKeyStore\"
+              location=\"safkeyring://IBMUSER/BOZRING\"
+              type=\"JCERACFKS\"
+              password=\"password\"/>
```
OnnoVdT: add comment saying keyring password must be "password", but is not actually used.

.setup/setup/setup-frontend-server.sh
```diff
@@ -62,56 +62,69 @@ fi
...
+    <logging traceSpecification=\"*=info\"
+             maxFileSize=\"20\"
+             maxFiles=\"10\" />
 
-    <!-- SSL Configuration (optional - for HTTPS) -->
-    <keyStore id="defaultKeyStore" password="Liberty" /> <!-- pragma: allowlist secret -->
+    <!-- SSL Configuration using RACF keyring -->
+    <ssl id=\"defaultSSLConfig\" keyStoreRef=\"defaultKeyStore\"/>
+    <keyStore id=\"defaultKeyStore\"
+              location=\"safkeyring://IBMUSER/BOZRING\"
```
OnnoVdT: use variables for userid and keyring

.setup/setup/zosconnect-ssl/tls.xml
```diff
@@ -0,0 +1,18 @@
...
+              type="JCERACFKS"
+              password="password"/>
```
OnnoVdT: add comment saying keyring password must be "password", but is not actually used.

.setup/setup-common.sh
OnnoVdT: Indeed strange to see bash here, as addcert.sh currently starts with #!/bin/sh

OnnoVdT (changes requested):

jbyibm:
This PR should be done on top of https://github.com/IBM/Bank-of-Z/pull/174.  #174 is what I consider now as production scripts. Most of the code before was done in an iterative manner to allow members to have a ZDVT as soon as possible.

The global section of `config.yaml` must have specific values for the targeted ZDVT.

See: https://github.com/IBM/Bank-of-Z/blob/2bb54711998ad42fd404318268a1edc6dbf05332/.setup/config/config.yaml#L10

jbyibm:
Global question:
- What is this zconfig folder:
  - https://github.ibm.com/IBMZSoftware/zconfig/blob/main/zconfig/src/zconfig/tasks/zosconnect/server.py

.setup/config/config.yaml
jbyibm: > @jbyibm please provide input here. Probably 'ASM.' needs the same update suggested for IMS

Must be in the global section of config.yaml now.

.setup/setup-common.sh
jbyibm: Bash is a must, and all scripts must start with  `#!/usr/bin/env bash`.

.setup/setup/addcert.sh
jbyibm: > should use the right env var - see also up in tls.xml - @jbyibm what is the right env var to reference here?

See config.yml

.setup/config/config.yaml
jbyibm: Not hardcoded anymore

.setup/config/config.yaml
QihanQiu1: Both replaced with variables. asm_lib_hlq: "ASM.SASMMOD1" → asm_hlq: "ASM" in global, referenced as {{ global.asm_hlq }}. ims_macro_lib: "IMSV15.SDFSMAC" → {{ ims.ims_sys_hlq }}.SDFSMAC

.setup/setup-common.sh
QihanQiu1: Kept bash as suggested by Jean-Yves for now.

.setup/setup/addcert.sh
QihanQiu1: I changed to userid=${ZOS_ADMIN_USER} — sourced from config.yaml via setenv.sh

.setup/setup/addcert.sh
QihanQiu1: > Some logic feels off here, line [352](https://github.com/IBM/Bank-of-Z/pull/165/files#diff-ee029e2a11478fe867924406448d9a5c2c2cf8b6545e09fb364fefd63151135eR350) print_success "RACF keyring and certificates created successfully" in setup-common.sh says it was successful in setup of RACF keyring and certs giving false understand that at this juncture, you could actually still fail to start Liberty without a cert. Feels like you should have bubbled up the error.

I have instead made the error message fatal, which should propagate to setup-common.sh causing it to exit with an error. setup-common.sh wraps the call with if bash .setup/setup/addcert.sh; then ... else print_error ... exit 1; fi.

QihanQiu1: > need to check return codes here. I think something called 'addcert' probably shouldn't be deleting a cert as a side effect - perhaps this should be broken into 2 scripts? At the start, should probably check that the user running has the authority to make these changes - and the user running should be the userid used - but I'd like @jbyibm 's opinion and @ddimatos 's on this.

I've made a new file clearcert with this logic.

.setup/setup/setup-frontend-server.sh
QihanQiu1: I've removed the python and instead used bash heredocs (cat > file << EOF)

.setup/setup/setup-zosconnect-server.sh
QihanQiu1: I've removed the python and instead used bash heredocs (cat > file << EOF)

.setup/setup/setup-zosconnect-server.sh
QihanQiu1: I've added a verify_file_written() helper that checks if the file is empty or missing.

src/frontend/server.js
QihanQiu1: This file seems to no longer exist so I removed it on my end as well.

.setup/config/config.yaml
QihanQiu1: It seemed that the port was already occupied by a different process, so I swapped the existing port values in order to resolve it.

.setup/jcl/ims/Ims-dynalloc.j2
QihanQiu1: I've made changes so that {{ ims.asm_lib_hlq }} is replaced with {{ ims.asm_hlq }}.SASMMOD1 and {{ ims.ims_macro_lib }} is replaced with {{ ims.ims_sys_hlq }}.SDFSMAC — HLQ as variable.

.setup/setup/zosconnect-ssl/http-endpoint.xml
QihanQiu1: I've removed this file entirely since it didn't seem necessary. This file should be written at setup time by setup-zosconnect-server.sh using the heredocs like ${ZOSCONNECT_HTTP_PORT}, ${ZOSCONNECT_HTTPS_PORT}, ${ZOS_ADMIN_USER}, and ${ZOS_KEYRING} from config.yaml.

.setup/setup/addcert.sh
QihanQiu1: I've made changes to replace it with RDATALIB.

.setup/setup/addcert.sh
QihanQiu1: I added a comment above the gencert-eku.sh call: "gencert-eku.sh removes any existing cert with label '$label' from the keyring before generating a fresh one."

.setup/setup/addcert.sh
QihanQiu1: I should have replaced all references to IBMUSER and BOZRING with ${ZOS_ADMIN_USER} and ${ZOS_KEYRING}.

.setup/setup/gencert-eku.sh
QihanQiu1: gencert-eku.sh now only generates and imports the cert. The RACDCERT CONNECT(LABEL('BoZ') RING(...) DEFAULT) step is in addcert.sh after gencert-eku.sh returns.

.setup/setup/gencert-eku.sh
QihanQiu1: The hardcoded JAVA_HOME=/usr/local/sandboxes/tools/J21.0_64 fallback has been removed. JAVA_HOME must be set in the environment (sourced from config.yaml via setenv.sh). If not set, the script exits with a FATAL error.

.setup/setup/gencert-eku.sh
QihanQiu1: I did switch the value to /usr/lpp/IBM/zoau but I added a fallback to /usr/lpp/IBM/zoautil since at least for me /usr/lpp/IBM/zoautil was the path that my image uses.

.setup/setup/gencert-eku.sh
QihanQiu1: Changed to TMPDIR=${TMPDIR:-/tmp}/boz-cert-$$

.setup/setup/gencert-eku.sh
QihanQiu1: Changed to mkdir -p "$TMPDIR" && chmod 700 "$TMPDIR".

.setup/setup/setup-frontend-server.sh
QihanQiu1: I removed the python so the tagging isn't needed anymore

.setup/setup/addcert.sh
QihanQiu1: ca_label=${ZOS_CA_LABEL} in both addcert.sh and gencert-eku.sh. ZOS_CA_LABEL is defined in config.yaml as zos_ca_label: "VSICA" and exported by setenv.sh

.setup/setup/addcert.sh
QihanQiu1: I made it so all connect operations now refresh only DIGTRING. DIGTCERT refresh only issued after RACDCERT GENCERT and RACDCERT ADD

.setup/setup/gencert-eku.sh
QihanQiu1: I replaced MAX_VALIDITY_MS = 397 days with the schedule as constants: DAYS_200 (until 2027-03-14), DAYS_100 (until 2029-03-14), DAYS_47 (after). A maxValidityMs() method picks the correct limit based on today's date automatically

.setup/setup/gencert-eku.sh
QihanQiu1: Comment added: # Exit immediately on any non-zero return code.

.setup/setup/gencert-eku.sh
QihanQiu1: Changed to ALLOC MOD CATALOG since it seems to overwrite

.setup/setup/setup-frontend-server.sh
QihanQiu1: Both added to httpEndpoint in the file

.setup/setup/setup-frontend-server.sh
QihanQiu1: sslProtocol="TLSv1.2,TLSv1.3" added to restrict to TLS 1.2+. enabledCiphers="-TLS_RSA_*" was also added initially but removed. I was getting an IllegalArgumentException: Unsupported CipherSuite: -TLS_RSA_* for negation syntax.

.setup/setup/setup-zosconnect-server.sh
QihanQiu1: Changed to ${CICS_HOST}, sourced from config.yaml (cics.host: "localhost") and exported by setenv.sh

.setup/setup/setup-zosconnect-server.sh
QihanQiu1: Comment added: "For JCERACFKS (SAF keyring) keystores, Liberty requires the password attribute to be present but ignores its value — 'password' is the conventional placeholder. The keyring itself is protected by RACF, not by this field."

.setup/setup/setup-frontend-server.sh
QihanQiu1: All four added to the frontend server.xml: httpDispatcher enableWelcomePage="false", config updateTrigger="disabled", applicationMonitor dropinsEnabled="false" updateTrigger="mbean", webContainer disableXPoweredBy="true".

updateTrigger="mbean" was chosen over "disabled" for applicationMonitor because Wazi Deploy drops new WARs into the apps/ directory after Liberty is already running. With "disabled", Liberty never rescans and the WARs are only picked up on a full restart. With "mbean", Liberty rescans only when explicitly triggered — no idle polling overhead, but the deploy task can signal a reload. task-wazi-deploy.sh issues an opercmd "C/S BAQ${APP_SHORT_NAME}" restart after each deploy to trigger this.

.setup/setup/gencert-eku.sh
QihanQiu1: RACDCERT DISENGAGE / cert removal block has been removed from gencert-eku.sh. Teardown (DELRING, cert deletion) is now handled exclusively by clearcert.sh

AndrewTwydell (commented):
I'm seeing a lot of references in here to the frontend and z/OS connect API sharing a liberty instance. This is not the case for the configuration in `main` - has the direction changed? 

I am of the understanding that we're deliberately NOT deploying the frontend to the z/OS connect server.

QihanQiu1:
> deliberately NOT deploying the frontend to the z/OS connect server.

My original approach was the two separate servers like we have in main and using CORS to allow cross-origin requests from the frontend origin and z/OS Connect side. However I ran into a lot of problems as I was trying to implement this, but it has been a while so I do not remember the exact problems I encountered. (Maybe something about how API calls were going to a different port and browsers treated it as a different origin so it required a lot of configurations behind the scenes?) I was experiencing difficulties trying to resolve these issues particularly with CORS so I fell back to a single Liberty instance serving both the frontend WAR at / and the API WAR at /api on the same origin and port instead. config.js then just uses a relative baseUrl: '/api' which works for HTTP and HTTPS automatically with no CORS involved. However, if the deliberate direction was to keep the two servers separate and my current implementation is bad practice,I am happy to take another look and jab at the task. Perhaps there was some working CORS configuration I missed that we had earlier? I would also love to learn some more about why the choice was made to deliberately not deploy the frontend to the z/OS connect server.

AndrewTwydell:
@QihanQiu1 Yes I believe it's bad practice and not advised by z/OS connect to deploy other applications to their server. We were doing this originally but have since moved to deploy the frontend to it's own liberty instance. 

The existing CORS setup tolerates different ports - we're wildcarding the host and specifying `http` and the frontend port. If you have issues after changing it to `https` and the secure port, let me know and we can have a look

QihanQiu1:
> @QihanQiu1 Yes I believe it's bad practice and not advised by z/OS connect to deploy other applications to their server. We were doing this originally but have since moved to deploy the frontend to it's own liberty instance.
> 
> The existing CORS setup tolerates different ports - we're wildcarding the host and specifying `http` and the frontend port. If you have issues after changing it to `https` and the secure port, let me know and we can have a look

I believe I've gotten it to work with CORS now, if you could take a look at it!

.setup/config/config.yaml
```diff
@@ -23,7 +23,7 @@ global:
   # Tools
   java_home: "{{ sandbox.root }}/tools/J21.0_64"
   python_home: "/usr/lpp/IBM/cyp/v3r14/pyz"
-  zoau_home: "/usr/lpp/IBM/zoautil"
+  zoau_home: "/usr/lpp/IBM/zoau"  # pre-1.3 ZVDT images may need "/usr/lpp/IBM/zoautil"
```
jbyibm: This is not the common place.

.setup/config/config.yaml
```diff
@@ -66,6 +68,7 @@ global:
   ims_disabled: "false"
   ims_sys_hlq: "IMSV15"
   ims_java_home: "/usr/lpp/ims/imsjava"
+  asm_hlq: "ASM"
```
jbyibm: Move this in # z/OS System section

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,81 @@
...
+ca_label=${ZOS_CA_LABEL}
+ring=${ZOS_KEYRING}
+label='BoZ'
+
+## FIXED ##
+profile=$userid.$ring.LST
+
+# Validate required environment variable
+if [[ -z "$userid" ]]; then
+  echo "[addcert] FATAL: ZOS_ADMIN_USER is not set" >&2
```
jbyibm: Use print_info, ... not echo.

.setup/setup/addcert.sh
```diff
@@ -0,0 +1,81 @@
...
+#   - Bouncy Castle re-signs adding EKU serverAuth + SANs + CAB Forum validity
+#   - RACDCERT ADD FORMAT(CERTDER) returns the re-signed cert to RACF
+# =============================================================================
+
+## CUSTOMIZE ##
+userid=${ZOS_ADMIN_USER}
+ca_label=${ZOS_CA_LABEL}
+ring=${ZOS_KEYRING}
+label='BoZ'
+
```
jbyibm, line 37: Add this at the very beginning of the shell script. (like other sh)

```
# =========================
# Source library scripts
# =========================
SCRIPTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source "$SCRIPTS_DIR/../config/setenv.sh"
exec > >(while IFS= read -r line; do
    line="${line%"${line##*[![:space:]]}"}"
    [[ -z "$line" ]] && continue
    printf "${CYAN}[ADD-CERTS]${NC} %s\n" "${line}" 2>/dev/null || true
done) 2>&1
```

.setup/setup/clearcert.sh
```diff
@@ -0,0 +1,52 @@
...
+# Called by setup-common.sh as a teardown step before addcert.sh.
+# All RACF commands are suppressed — failures are expected on a clean system.
+# =============================================================================
+
+## CUSTOMIZE ##
+userid=${ZOS_ADMIN_USER}
+ca_label=${ZOS_CA_LABEL}
+ring=${ZOS_KEYRING}
+label='BoZ'
+
```
jbyibm, line 28: Same comment as .sh ^^

.setup/setup/gencert-eku.sh
```diff
@@ -0,0 +1,361 @@
...
+#      re-signs it with VSICA's key (exported as PKCS12DER, deleted after).
+#   4. dcp the new DER cert back to a RACF dataset.
+#   5. RACDCERT ADD FORMAT(CERTDER) — RACF matches the new cert to the
+#      private key it already holds (same public key).
+#   6. Connect cert to ${ZOS_KEYRING} as DEFAULT.
+#   7. Liberty uses safkeyring://${ZOS_ADMIN_USER}/${ZOS_KEYRING} (JCERACFKS).
+#
+# Called by addcert.sh after the keyring scaffold is in place.
+# =============================================================================
+
```
jbyibm, line 23: Same comment as .sh ^^

.setup/tasks/task-wazi-deploy.sh
```diff
@@ -217,6 +217,34 @@ fi
...
+print_info "Restarting FE${APP_SHORT_NAME} to load newly deployed frontend WAR..."
+opercmd "C FE${APP_SHORT_NAME}" 2>/dev/null || true
+sleep 8
+if [[ "$FRONTEND_SYS_PROCLIB" != "${APP_HLQ}.PROCLIB" ]]; then
+    opercmd "S FE${APP_SHORT_NAME}" 2>/dev/null || true
+else
+    jsub "${FRONTEND_SYS_PROCLIB}(FE${APP_SHORT_NAME}J)" 2>/dev/null || true
+fi
+print_success "FE${APP_SHORT_NAME} restart issued - server will be ready in ~20 seconds"
+
```
jbyibm, line 247: Why this ? 
We refresh the config in Wazi Deploy. This is not a good practice from a deployment point of view.

jbyibm (changes requested):
Some comments

.setup/config/config.yaml
QihanQiu1: Is zoautil the common place? Onno commented that ZOAU 1.4 (and it looks like 1.3 as well) switched to default path /usr/lpp/IBM/zoau so that was the reason I made the change.

OnnoVdT: https://www.ibm.com/docs/en/SSKFYE_1.3.x/zoau-program-directory/hal5130.pdf
/usr/lpp/IBM/zoau/v1r3
https://www.ibm.com/docs/en/SSKFYE_1.4.x/zoau-program-directory/hal5140.pdf
/usr/lpp/IBM/zoau/v1r4

jbyibm: so should be /usr/lpp/IBM/zoau/v1r4 @OnnoVdT

.setup/setup/addcert.sh
QihanQiu1: Done!

.setup/setup/addcert.sh
QihanQiu1: I changed my echo usage into prints.

.setup/config/config.yaml
QihanQiu1: Done.

.setup/config/config.yaml
QihanQiu1: I tried to update it to /usr/lpp/IBM/zoau/v1r4, but this broke the validation script since the ZVDT image we're running on has ZOAU installed at /usr/lpp/IBM/zoau without the version suffix, so I kept it as zoau_home: "/usr/lpp/IBM/zoau" for now.

.setup/tasks/task-wazi-deploy.sh
QihanQiu1: During testing, Wazi Deploy completed successfully but the new WAR wasn't being served and the old version kept running, so I added the restart blocks as a workaround. I think the issue might have been that: dropinsEnabled="false" in the server config was preventing configDropins/overrides/api.xml from loading at all and updateTrigger="disabled" in the frontend server.xml meant Liberty ignored refresh commands entirely. I was doing some testing to try to remove these manual refreshes, but encountering issues on this end still.

jbyibm: Today it works on the main branch. So what is your use case?
What is the difference?

jbyibm: I would say for now, commit with your code and create a GitHub issue for me. I'll look later at why you need to restart servers.

jbyibm (approved):
Looks Good

OnnoVdT (approved):
didn't look at everything again, but it's good for me

jbyibm:
Hello @ddimatos can you check your requested changes Thanks

.setup/setup-common.sh
ddimatos: Would not have been my choice to require bash, but if you say so :), the niceties bash provides are indeed 'nice' but most of the time easily aligned to more portable options.

.setup/setup/addcert.sh
ddimatos: Thank you, sorry for the late review ( I was traveling).

ddimatos:
@jbyibm , @QihanQiu1 @MikeFultonDev , I realize it is merged now but do want to confirm I did look over all the changes, I was traveling for some time and not able to have a look.

All looks good, lots of great feedback, thanks for including me.

ddimatos (commented):
LGTM

PR #179, opened 2026-07-23
Upgrade to zconfig 0.7.0

AndrewTwydell:
ZConfig is releasing 0.7.0 today - this has breaking changes so these are required updates for 0.7.0.


adn1107 (approved):
OK

PR #189, opened 2026-08-10
Fix issues occurring when using an account other than IBMUSER

jbyibm:
IBMUSER often has elevated or default privileges on z/OS systems, so operations that work seamlessly under that account may fail under a different user ID due to missing RACF authorizations. To fix this:

- Identify the required resources: Determine which datasets, RACF classes (e.g., FACILITY, SURROGAT), or system resources the process needs to access.
- Grant explicit permissions: Use RACF commands (e.g., PERMIT, ADDSD, ALTUSER) to grant the non-IBMUSER account the necessary access rights, rather than relying on implicit/default authority.
- Check UACC and group memberships: Ensure the account belongs to the correct RACF groups, and that Universal Access (UACC) settings don't silently block access.
- Review started task/APF authorization: If the process runs as a started task, verify the associated RACF STARTED class definition and that the user ID tied to the task has the right profile.
- Audit with SETROPTS/RACF reports: Use SETROPTS AUDIT or RACF reporting tools to trace access denials (ICH408I messages) and correct the specific missing permissions.
- Allow to use existing certificates


jbyibm:
Hello @QihanQiu1  
I did some updates in `.setup/setup/grant-db2-user.sh` that you will need to integrate into your current dev branch.

.setup/build/groovy/VanillaFrontend.groovy
```diff
@@ -144,9 +145,12 @@ try {
             def content = fileToModify.getText('UTF-8')
             // Replace only standalone occurrences of 9080 (word boundaries)
             content = content.replaceAll(/\b9080\b/, zosConnectHttpPort.toString())
+            content = content.replaceAll(/\9444\b/, zosConnectHttpsPort.toString())
```
QihanQiu1: I think this should be \b9444\b/

QihanQiu1 (commented):
The rest of the HTTPS related changes LGTM.

QihanQiu1 (approved):
LGTM

PR #191, opened 2026-08-10
Rename grant-db2-user.sh to grant-perm-user.sh and add RACF keyring grants for non-IBMUSER support

QihanQiu1:
- grant-db2-user.sh → grant-perm-user.sh — renamed to reflect its broader purpose beyond DB2
- Added RDATALIB keyring grants — the script now grants the target user CONTROL access to the RACF keyring (BOZRING) via the RDATALIB class, so Liberty can read the TLS certificate during HTTPS setup. Previously, only IBMUSER had this access.
- addcert.sh — RDATALIB commands in Step 3 made best-effort (|| true) so non-admin users don't abort if the profile is already defined by the admin via grant-perm-user.sh
- clearcert.sh — SETROPTS GENERIC/CLASSACT(RDATALIB) made best-effort (|| true) since these require RACF SPECIAL authority and may already be active
- Docs updated — environment-configuration.md, docs/backup/README.md updated to reference grant-perm-user.sh instead of the old manual JCL approach


jbyibm (approved):
Looks Good

PR #200, opened 2026-08-13
Refine IMS definition to make it more generic

jbyibm:
This PR externalizes hard-coded values for IMS to allow installation on different versions of IMS and/or ZDVT.


dennis-behm (approved):
@jbyibm successful install and deployment of your branch on my system.. This looks good to me

Issue #182, opened 2026-07-29
[Bug] Investigate why MODIFY REFRESH,APPS is insufficient for WAR reload on FEBOZ/BAQBOZ


QihanQiu1:
During testing in PR #165, Wazi Deploy's REFRESH,APPS completed successfully but the new WAR was not being served after deploy. Manual server restarts were added as a workaround in task-wazi-deploy.sh.

Possible root causes identified but not fully verified:
- dropinsEnabled=false in the z/OS Connect server template prevented configDropins/overrides/api.xml from loading
- updateTrigger=disabled in the frontend server.xml meant Liberty ignored refresh commands entirely

On main branch the refresh works. Need to understand what is different in this PR's two-server design (BAQBOZ for API, FEBOZ for frontend) that causes refresh to be insufficient.

Issue #202, opened 2026-08-13
[Feature] Provide the ability start/stop BOZ tasks/jobs


jbyibm:
Today's the tasks or jobs are started during the setup. We should provide a consistent way to allow starting tasks at IPL time and/or globally with start and stop dedicated scripts.

jbyibm:
@MikeFultonDev @dennis-behm  FYI

jbyibm:
During the development of this issue, we need to invest more in why we need to restart z/OS Connect and Fronted at the end of the setup.
- https://github.com/IBM/Bank-of-Z/blob/238ece3c153f395adfb37ef7b7c64dc40de0ce3d/.setup/setup-common.sh#L604

Wazi Deploy refresh system command should be enough. To see with the z/OS Connect team.

@dennis-behm FYI

PR #208, opened 2026-08-17
[Bug] Investigate why MODIFY REFRESH,APPS is insufficient for WAR reload on FEBOZ/BAQBOZ

jbyibm:
Fix for: https://github.com/IBM/Bank-of-Z/issues/182

### DIFF
--- a/.setup/setup-common.sh
+++ b/.setup/setup-common.sh
@@ -522,16 +522,14 @@ main_setup() {
         stage_setup_ims_bankz_regions
     fi
 
-    stage_setup_zosconnect_server
-
-    stage_setup_frontend_server
-
-    # Certificates run last so addcert.sh + gencert-eku.sh regenerate the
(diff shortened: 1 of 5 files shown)


.setup/setup-common.sh
```diff
@@ -602,6 +600,21 @@ main() {
             if [[ "$IMS_DISABLED" != "true" ]]; then
                 stage_populate_ims_database
             fi
+
```
dennis-behm, line 603: Can we probably turn that into its own **stage** and "subcommand" to re-start frontend servers?
See also - https://github.com/IBM/Bank-of-Z/issues/202

jbyibm: Yes, but when we will implement #202. 
#202. will require more time to review the overall code.

dennis-behm (approved):

PR #209, opened 2026-08-17
Installation verification routines

dennis-behm:
This expands the `setup-common` install process with a set of tests checking the availability of the application for both backend implementations. It can be used stand-alone as well for the PR workflow.


jbyibm (approved):
Looks good

PR #216, opened 2026-08-25
ims db2 zosconnect deployment diagram updated

swathi-sadananda:
Updated ims db2 zosconnect deployment diagram  in the Architecture section


adn1107 (approved):
LGTM

PR #219, opened 2026-08-27
Feature/simplify user build setup

dennis-behm:
This adopts the conditional logic to address the bug in StartInit, when the directories are not present


jbyibm (approved):
Looks good for me.

PR #224, opened 2026-08-31
A bugfix for EQAPROF and IMS Debugging tutorial

akarputkin:
This PR adds a tutorial for Debugging an IMS Transaction using VS Code and IMS Transaction Isolation Profile;
Also it fixes an issue with EQAPROF, it is unable to work with Java from tools.tar as it doesn't have the program control bit on. I switched it back to /usr/lpp/java


MikeFultonDev:
Are there other tools that need the tools.tar version of Java? Presumably there are - if not we should remove it from the tools.tar. Assuming there is, it sounds like the 'install' from tools.tar isn't correct - there should be a post-install step to set the APF bits on appropriately OR we could use 'pax' if the original has the APF bits on so that they are preserved. If we want to use pax, Anthony Giorgio from ZOAU knows how to do this since ZOAU also has APF auth bits.

dennis-behm:
a fix of the (internal) tools install is underway. The updated java is required for the gradle builds under z/OS Unix System services.

MikeFultonDev (approved):
LGTM

PR #226, opened 2026-09-01
Added Batch Demo latest programs

reachsenthilnathan:


reachsenthilnathan:
@MikeFultonDev  here is the PR with the latest 3 Batch programs

MikeFultonDev (commented):
the target should be BANKZ.V0R1M0.LOAD, not LOADLIB - can you change?
The report DD should be MOD and not NEW so that you can run multiple times.

MikeFultonDev (approved):
LGTM

PR #237, opened 2026-09-10
Reduce time to prepare the environment file

chrishodgins:
## Summary

(Generated with help from IBM Bob)

Improve `setenv.sh` performance by replacing per-variable `get_section_value` shell subprocesses with a single Jinja2 template rendering pass via Python.

## Problem

The previous implementation called `get_section_value` (a Python subprocess invocation) individually for every variable in `.env` — roughly 60+ calls per setup run. Each call incurs Python startup overhead, making environment file generation slow.

## Solution

The `.env` file is now generated using a two-step approach:

1. A static `.env.template` file is written using a heredoc with `'EOF'` (no shell expansion), containing Jinja2-style `{{ section.key }}` placeholders.
2. A single `python3 config.py --resolve-template` call reads the `.env.template` and configuration yaml file, rendering the template in one pass, and writes the final `.env` file.

The template file is cleaned up immediately after rendering.

## Changes

- **`.setup/config/setenv.sh`**: Replaced all `$(get_section_value ...)` calls with `{{ section.key }}` Jinja2 placeholders in a quoted heredoc. Added a single `python3 config.py --resolve-template` call to render the template.
- **`.setup/lib/config.py`**: Added `resolve_template()` function and `--resolve-template <file>` CLI mode. Moved `load_config`/`render_config` calls before argument dispatch so the config is loaded once regardless of mode.

## Notes

- Also fixes a minor inconsistency: `DBB_LOG_FOLDER` was previously set twice (direct + `:-` fallback). The duplicate assignment is removed, keeping only the `:-` form.
- `SCAN_PYENV_ACTIVATE_PATH` was previously using `$PYENV_ACTIVATE_PATH` as its default variable name (likely a bug); corrected to `$SCAN_PYENV_ACTIVATE_PATH`.

### DIFF
--- a/.setup/config/setenv.sh
+++ b/.setup/config/setenv.sh
@@ -32,12 +32,14 @@ fi
 set -e
 
 
+TEMPLATE_FILE="${LOCAL_SCRIPTS_DIR}/.env.template"
+
 if [[ ! -f "$ENV_FILE" || "$ENV_FILE" -ot "$CONFIG_FILE" || "$ENV_FILE" -ot "${BASH_SOURCE[0]}" ]]; then
     print_warning "Creating $ENV_FILE file ..."
     print_warning " - Not already exists or not in sync with:"
     print_warning "   - '$CONFIG_FILE'"
     print_warning "   - '${BASH_SOURCE[0]}'"
-    cat > "$ENV_FILE" <<EOF
+    cat > "$TEMPLATE_FILE" <<'EOF'
 # =========================
 # Environment
 # =========================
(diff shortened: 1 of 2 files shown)


dennis-behm (commented):
When do you see the script being continuously invoked?

Usually once the repo is cloned, and the `setenv.sh` ran once, subsequent invocations sourcing the .env file. It looks like this:

```
.setup/setup-common.sh verify-installation
[INFO] All variables are properly set.

=================================================================
= Detecting Bank of Z location...
....
```

However, if you clone the repo new, it will re-run it.

dennis-behm (commented):
@chrishodgins I tested these updates. These look good to me! 

On my fork of bank-of-z, I have enabled a GitHub actions pipeline, which clones the repo on every pipeline run and I noticed the same behaviour, that it kept taking about 1 minute to setup this. See https://github.com/dennis-behm/Bank-of-Z/actions/runs/34579885327/job/103200817717

dennis-behm (approved):

chrishodgins:
@dennis-behm not continuously invoked, but every config.yaml change forces the .env rewrite. So if you are setting up for the first time,(like I was) it can be frustrating to wait 3 minutes before you can validate the pre-reqs.

Thanks for the review. The pipeline run is currently showing failed due to a logs issue which looks unrelated.

PR #238, opened 2026-09-10
Use THRU to exit the para as GO TO has limitation to go back to paragraph

PrasangAPrajapati:
https://www.mainframemaster.com/tutorials/cobol/procedure-branching 

image (7)


MikeFultonDev (approved):
approved. We will get Venkatesh to also review this next week and we can add an additional PR at that time if required.

Issue #239, opened 2026-09-11
[Editors] Ability to resolve derived BMS copybooks


dennis-behm:
Editors are not yet configured to checkout the remote library for generated BMS copybooks. 

Adding a property group definition to `zapp.yaml` including a remote sys lib, will fix this.

PR #241, opened 2026-09-11
Add remote MVS library configuration to zapp.yaml for resolving BMS copybooks

dennis-behm:
Implements #239


jbyibm (approved):
looks Good.

PR #242, opened 2026-09-11
Prevent early script exit while flushing stdout and stderr

chrishodgins:
## Summary

Adds an `EXIT` trap (`trap 'exec >&- 2>&-; wait' EXIT`) across setup and task bash scripts to ensure all background process substitutions redirecting standard output and standard error finish flushing and cleanly exit before the main script process terminates.

## Problem

Scripts redirecting `stdout` and `stderr` via process substitution (e.g., `exec > >(while IFS= read -r line; do ... done) 2>&1`) can terminate before the subshell completes processing and writing log buffers. This can lead to truncated or dropped log outputs, race conditions on script completion, and intermittent early termination issues.

## Solution

Added `trap 'exec >&- 2>&-; wait' EXIT` after process substitution setup across:
- **Setup scripts** under `.setup/setup/`
- **Task scripts** under `.setup/tasks/`

Closing file descriptors (`exec >&- 2>&-`) sends EOF to the background logging loop, and `wait` ensures the subshell finishes draining before the parent process exits.

## Validation

- Verified that all modified scripts retain proper syntax and structure.
- Confirmed trap behavior correctly flushes outputs on normal script termination and exits cleanly.

## Testing

- Manual testing of a grub client install of Bank of Z
- Manual testing of a local USS based run of `setup-common.sh validate-prereqs`

### DIFF
--- a/.setup/setup/addcert.sh
+++ b/.setup/setup/addcert.sh
@@ -28,6 +28,7 @@ exec > >(while IFS= read -r line; do
     [[ -z "$line" ]] && continue
     printf "${CYAN}[ADD-CERTS]${NC} %s\n" "${line}" 2>/dev/null || true
 done) 2>&1
+trap 'exec >&- 2>&-; wait' EXIT
 
 finalize_results() {
     RC=$?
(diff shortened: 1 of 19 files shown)


jbyibm (approved):
Looks Good

PR #243, opened 2026-09-11
bank-of-z server stop and start scripts

dennis-behm:
# Summary

Refactors server lifecycle management for the Bank of Z application by extracting stop/start logic into dedicated, reusable scripts. Previously, server stop commands were inlined directly in `setup-common.sh`. This PR introduces `servers-stop.sh` and `servers-start.sh` as standalone scripts with scoped execution support (`--all`, `--ims-only`, `--cics-only`, `--frontend-only`), and adds a top-level `runtime-manage.sh` orchestrator for managing the full server lifecycle (stop, start, restart) independently of provisioning or data operations.

# Files Changed

📄 `.setup/runtime-manage.sh`

New top-level orchestration script for managing the Bank of Z server lifecycle on z/OS USS. Accepts an `action` (`stop`, `start`, `restart`) and an optional `scope` flag, then delegates to `servers-stop.sh` and/or `servers-start.sh` accordingly. Includes usage documentation, input validation, and clear stage/success messaging. Intended for use when the z/OS infrastructure is restarted and servers need to be brought back up manually without touching any datasets or application data.

📄 `.setup/setup-common.sh`

Refactored the `stage_stop_tasks` function to remove all inlined server stop commands (IMS regions, CICS, z/OS Connect). These are now delegated to the new `servers-stop.sh --all` script. The only remaining inline command is the legacy `jcan P "IMS1*"` call, which has been updated to use `|| true` for safer error handling. Also adds a guard to exit early with an error if `servers-stop.sh` is not found.

📄 `.setup/setup/servers-start.sh`

New script responsible for starting Bank of Z runtime servers. Implements four scoped start functions — `start_ims_control`, `start_ims_regions`, `start_cics`, and `start_frontend` — each handling the appropriate `opercmd`/`jsub` calls with correct startup ordering and delays. Supports the `IMS_DISABLED` environment variable to skip IMS tasks entirely. Start order when using `--all` is IMS → CICS → Frontend.

📄 `.setup/setup/servers-stop.sh`

New script responsible for stopping Bank of Z runtime servers without modifying any datasets or application data. Implements four scoped stop functions — `stop_frontend`, `stop_cics`, `stop_ims_regions`, and `stop_ims_control` — with proper teardown ordering and safe error handling via `|| true`. Stale JCL stop members are removed before resubmission to prevent outdated JCL from executing. Stop order when using `--all` is Frontend → CICS → IMS. Supports the `IMS_DISABLED` environment variable to skip IMS tasks.

### DIFF
--- a/.setup/runtime-manage.sh
+++ b/.setup/runtime-manage.sh
@@ -1,0 +1,152 @@
+#!/usr/bin/env bash
+
+#########################################################
+# Runtime Management Script for Bank of Z
+# This script runs directly on z/OS USS (not remotely)
+#
+# Manages the lifecycle (stop / start / restart) of the
+# Bank of Z runtime servers WITHOUT touching any datasets
+# or application data.
+#
+# Used when the z/OS infrastructure is restarted and the
+# Bank of Z tasks need to be brought back up manually.
+#
+# Delegates to:
+#   .setup/setup/servers-stop.sh
+#   .setup/setup/servers-start.sh
+#
+# Usage:
+#   bash runtime-manage.sh <action> [scope]
+#
+# Actions:
+#   stop      Stop all servers (no data deletion)
+#   start     Start all servers
+#   restart   Stop then start all servers
+#
+# Scope (optional, default: --all):
+#   --all           IMS + CICS + z/OS Connect + Frontend
+#   --ims-only      IMS control tasks + IRLM + app regions only
+#   --cics-only     CICS region only
+#   --frontend-only z/OS Connect + Frontend Liberty only
+#
+# Examples:
+#   bash runtime-manage.sh stop
+#   bash runtime-manage.sh start --ims-only
+#   bash runtime-manage.sh restart
+#########################################################
+
+set -e
+
+# =========================
+# Source library scripts
+# =========================
+SCRIPTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPTS_DIR/config/setenv.sh"
+
+#########################################################
+# Usage
+#########################################################
+print_usage() {
+    echo "Usage: bash runtime-manage.sh <action> [scope]"
+    echo ""
+    echo "Actions:"
+    echo "  stop      Stop all servers (no data deletion)"
+    echo "  start     Start all servers"
+    echo "  restart   Stop then start all servers"
+    echo ""
+    echo "Scope (optional, default: --all):"
+    echo "  --all           IMS + CICS + z/OS Connect + Frontend  (default)"
+    echo "  --ims-only      IMS control tasks + IRLM + app regions only"
+    echo "  --cics-only     CICS region only"
+    echo "  --frontend-only z/OS Connect + Frontend Liberty only"
+    echo ""
+    echo "Examples:"
+    echo "  bash runtime-manage.sh stop"
+    echo "  bash runtime-manage.sh start --ims-only"
+    echo "  bash runtime-manage.sh restart"
+    echo ""
+    echo "Environment variables:"
+    echo "  IMS_DISABLED   Set to true to skip all IMS tasks (default: false)"
+}
+
+#########################################################
+# Main
+#########################################################
+main() {
+    local action="${1:-}"
+    local scope_flag="${2:---all}"
+
+    # Validate scope
+    case "${scope_flag#--}" in
+        all|ims-only|cics-only|frontend-only) ;;
+        *)
+            print_error "Unknown scope: ${scope_flag}"
+            echo ""
+            print_usage
+            exit 1
+            ;;
+    esac
+
+    # Detect repo location (sets BANK_DIR, EXECUTION_MODE)
+    detect_bank_of_z_location
+
+    local stop_script="$BANK_DIR/.setup/setup/servers-stop.sh"
+    local start_script="$BANK_DIR/.setup/setup/servers-start.sh"
+
+    case "$action" in
+        stop)
+            print_stage "ACTION: Stop Bank of Z servers (scope: ${scope_flag#--})"
+            if [ ! -f "$stop_script" ]; then
+                print_error "Stop script not found: $stop_script"
+                exit 1
+            fi
+            print_info "Executing: bash $stop_script $scope_flag"
+            bash "$stop_script" "$scope_flag"
+            print_stage "STOP COMPLETE"
+            print_success "All requested servers have been stopped."
+            ;;
+        start)
+            print_stage "ACTION: Start Bank of Z servers (scope: ${scope_flag#--})"
+            if [ ! -f "$start_script" ]; then
+                print_error "Start script not found: $start_script"
+                exit 1
+            fi
+            print_info "Executing: bash $start_script $scope_flag"
+            bash "$start_script" "$scope_flag"
+            print_stage "START COMPLETE"
+            print_success "All requested servers have been started."
+            ;;
+        restart)
+            print_stage "ACTION: Restart Bank of Z servers (scope: ${scope_flag#--})"
+            if [ ! -f "$stop_script" ]; then
+                print_error "Stop script not found: $stop_script"
+                exit 1
+            fi
+            if [ ! -f "$start_script" ]; then
+                print_error "Start script not found: $start_script"
+                exit 1
+            fi
+            print_info "Executing: bash $stop_script $scope_flag"
+            bash "$stop_script" "$scope_flag"
+            print_info "Executing: bash $start_script $scope_flag"
+            bash "$start_script" "$scope_flag"
+            print_stage "RESTART COMPLETE"
+            print_success "All requested servers have been restarted."
+            ;;
+        -h|--help|help|"")
+            print_usage
+            ;;
+        *)
+            print_error "Unknown action: $action"
+            echo ""
+            print_usage
+            exit 1
+            ;;
+    esac
+}
+
+# Run main function
+main "$@"
+exit $?
+
+# Made with Bob
--- a/.setup/setup-common.sh
+++ b/.setup/setup-common.sh
@@ -25,54 +25,25 @@ source "$SCRIPTS_DIR/config/setenv.sh"
 stage_stop_tasks() {
     set +e
     print_stage "STAGE: Stop Bank of Z running tasks (if any)"
+
     # =========================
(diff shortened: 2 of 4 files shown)


chrishodgins:
Thoughts so far:
- CICS shutdown should be `/F CICSBOZ,CEMT PERFORM SHUTDOWN`. This asks the CEMT transaction to shutdown CICS cleanly. The current `/P CICSBOZ` isn't a clean shutdown.
- We can already do a .setup/setup-common.sh verify-installation which sounds similar to what you are suggesting with a runtime health-check.
- IMS startup isn't quite right but you've mentioned that. Not quite connecting back properly yet.
- Restart brought everything down without any problems. It even removed the duplicate IMS2JMP1 job that always seems to appear during setup.
- Maybe nicer to have a single scope command rather than arguments. For example what does `runtime-manage.sh start --ims-only --cics-only` really do anyway. Plus the -only bit doesn't add much. Maybe just `runtime-manage.sh start cics|ims|frontend|all` ?
- Starting up individually also worked well (ignoring the IMS thing).
- No support for running that command from grub remotely? 
- I just spotted that start/stop default to `--all` . Might be better to force the user to be explicit about they want. Should stop without options really bring down your bank?


Overall, worked great. We could more generally refine the front-end shell experience a little more for boz but this fits in nicely to what already exists.

chrishodgins:
IMS console commands to potentially bring IMS down cleanly. Following the guide here, I'm not an IMS SME.
https://www.ibm.com/docs/en/ims/15.4.0?topic=automation-shutting-down-ims

```
# Shutdown the dependent regions
/IMS2STOP REGION JOBNAME IMS2JMP1
/IMS2STOP REGION JOBNAME IMS2MPP1
/IMS2STOP REGION JOBNAME IMS2MPP2

# Shutdown the CTL region safely (this is the key bit I think)
/IMS2CHECKPOINT PURGE

# We didn't have IRLM up and I can't recall whether it was there initially
# Not tried this one as my IRLM wasn't there
/F IRLM,ABEND,NODUMP

# Shutdown the IMSPLEX, this covers a few address spaces
/F IMS2SCI,SHUTDOWN CSLPLEX

# Shutdown IMS Connect (not sure about the order of this one)
/F IMS2HWS,SHUTDOWN MEMBER
```

jbyibm:
Hello @dennis-behm 
Can you consider this when the setup user can't install PROC at the system level?
- https://github.com/IBM/Bank-of-Z/blob/d00876b8f2a3e757457d8d35302bb196e2f73c31/.setup/setup/setup-frontend-server.sh#L250

dennis-behm:
### Summary of Recent Updates to Server Lifecycle Management

Consolidated the runtime server start, stop, and restart routines directly into [`.setup/runtime-manage.sh`](.setup/runtime-manage.sh):

1. **IMS Lifecycle Consolidation**:
   - Integrated the start/stop jobs from [`setup-ims-bankz-regions.sh`](.setup/setup/setup-ims-bankz-regions.sh) into `runtime-manage.sh` (`stop_ims_regions`, `stop_ims_control`, `start_ims_control`, `start_ims_regions`).
   - Removed destructive dataset member cleanup (`mrm`) so stop/start runs idempotently without removing required PDS members.
   - Added IMS post-start verification (`verify_ims`) to confirm IMS Control Region, IMS Connect (`${IMS_DATASTORE}HWS`), and port listener status via `opercmd` and `netstat`.

2. **USS Portability**:
   - Replaced non-standard `grep -q` calls with USS-compatible redirection `grep ... >/dev/null 2>&1` across all server verification checks.

dennis-behm:
@swathi-sadananda, do you know if the new runtime-manage.md under docs will automatically appear on published GitHub pages?

swathi-sadananda:
> @swathi-sadananda, do you know if the new runtime-manage.md under docs will automatically appear on published GitHub pages?

@dennis-behm I don't see this file in the docs folder. If it's available and mapped in the TOC, it should show up on the GitHub site after the merge.

swathi-sadananda:
> > @swathi-sadananda, do you know if the new runtime-manage.md under docs will automatically appear on published GitHub pages?
> 
> @dennis-behm I don't see this file in the docs folder. If it's available and mapped in the TOC, it should show up on the GitHub site after the merge.

@dennis-behm Otherwise, you can share it with me, and I'll update it

.setup/runtime-manage.sh
```diff
@@ -0,0 +1,562 @@
...
+
+#########################################################
+# Stop CICS region
+#########################################################
+stop_cics() {
+    print_stage "STAGE: Stop CICS region"
+    set +e
+
+    # Check whether the CICS region is currently active
+    if opercmd "D A,CICS${APP_SHORT_NAME}" 2>/dev/null | grep "CICS${APP_SHORT_NAME}" >/dev/null 2>&1; then
```
chrishodgins: For the moment, this is likely fine but we should consider whether we'd want to revisit this later so the normal shutdown waits for the job to exit but there is a force stop option if we really need to shut things down.

.setup/runtime-manage.sh
```diff
@@ -0,0 +1,562 @@
...
+
+    # Resolve the current WTOR reply ID for the CTL region so we can
+    # issue /STOP REGION commands through the IMS console.
+    local REPLID
+    REPLID=$(opercmd "D R,JOB=${IMS_DATASTORE}CTL" 2>/dev/null \
+             | grep -i "IMS READY" | awk '{print $1}' | head -1)
+
+    if [[ -n "$REPLID" ]]; then
+        print_info "Stopping dependent regions via IMS console (REPLID=${REPLID})..."
+        opercmd "${REPLID},/STOP REGION JOBNAME ${IMS_DATASTORE}JMP1" 2>/dev/null || true
```
chrishodgins: Do we need the sleeps between the opercmds here? We could issue them all at once and then wait?

.setup/runtime-manage.sh
```diff
@@ -0,0 +1,562 @@
...
+        print_warning "CTL WTOR reply ID not found - skipping /CHECKPOINT PURGE"
+    fi
+
+    # Step 2: Graceful IMS Connect shutdown
+    print_info "Shutting down ${IMS_DATASTORE}HWS (IMS Connect) gracefully..."
+    opercmd "F ${IMS_DATASTORE}HWS,SHUTDOWN MEMBER" 2>/dev/null || true
+    sleep 3
+
+    # Step 3: Stop ODBM and DRD (no graceful IMS console command for these)
+    print_info "Stopping ${IMS_DATASTORE}ODB..."
```
chrishodgins: When I've run this manually in the past, I didn't have to specifically cancel the ODD/DRC/CTL regions. 

From a quick test:
- CTL/DRC shutdown during the `/CHECKPOINT PURGE`
- ODB/RM/SCI all shutdown together during the `F ${IMS_DATASTORE}SCI,SHUTDOWN CSLPLEX`

.setup/runtime-manage.sh
```diff
@@ -0,0 +1,562 @@
...
+
+    print_info "Starting ${IMS_DATASTORE}OM..."
+    opercmd "S ${IMS_DATASTORE}OM" 2>/dev/null || true
+    sleep 2
+
+    print_info "Starting ${IMS_DATASTORE}RM..."
+    opercmd "S ${IMS_DATASTORE}RM" 2>/dev/null || true
+    sleep 2
+
+    print_info "Submitting ${IMS_APP_HLQ}.PROCLIB(${IMS_DATASTORE}CTL) via jsub..."
```
chrishodgins: I see CICS and z/OS Connect both check their proclib against the APP_HLQ proclib. Is that required for IMS as well?

.setup/runtime-manage.sh
```diff
@@ -0,0 +1,562 @@
...
+        frontend)
+            stop_frontend
+            ;;
+    esac
+}
+
+#########################################################
+# Dispatch start by scope
+#########################################################
+do_start() {
```
chrishodgins: General point about the start_* paths. We need to add checks to make sure these address spaces aren't already up and running before starting them. I managed to start IMS address spaces twice. It did manage to shut it all down again but for a while I have multiple instances of the same IMS address spaces.

chrishodgins: Ugly example of this. I started CICS twice using `.setup/runtime-manage.sh start cics`, so ended up with one active CICS and a duplicate CICS job waiting to run. I then used the `.setup/runtime-manage.sh stop cics`. This found the CICS region still active, then cancelled the duplicate job, leaving CICS up and running.

.setup/runtime-manage.sh
chrishodgins: From testing this out, it always believes CICS is active.

dennis-behm: Thanks for testing this thoroughly. Working on addressing your feedback.

dennis-behm:
### Updates added today (2026-09-24)

Four commits landed today on `.setup/runtime-manage.sh`, all addressing feedback from earlier reviews:

1. **Idempotent start/stop guards** (`617fe4f`) — All start functions now check whether a task is already running before issuing a start command, and all stop functions validate that each task has actually stopped after the shutdown sequence. A final `D A,*` display is issued when any IMS task remains active.

2. **Graceful shutdown wait periods** (`ae1a074`) — Added timed polling loops (up to 25–30 s) across all stop functions (`stop_frontend`, `stop_cics`, `stop_ims_regions`, `stop_ims_control`) so that the script waits for a clean shutdown before issuing a `C` cancel as a last resort.

3. **Post-start verification** (`4f9dc23`) — `start_ims_control` and `start_ims_regions` now call the shared `wait_for_task_running` helper after starting each task (IRLM, SCI, OM, RM, CTL, ODB, HWS, MPP1, MPP2, JMP1) to confirm startup rather than relying on fixed sleep delays.

4. **`verify` action** (`31567b3`) — Added a new top-level `verify` action (`bash runtime-manage.sh verify all|ims|cics|frontend`) that checks the running state of every task in scope and reports pass/fail without making any changes. This addresses the earlier comment about runtime health-checking being a useful standalone capability.

.setup/runtime-manage.sh
dennis-behm: Implemented a check if tasks are already running

.setup/runtime-manage.sh
dennis-behm: Addressed in today's update

PR #248, opened 2026-09-21
Simple setup-common script output logging

dennis-behm:
# Summary

Introduces a logging setup function to the common setup script that automatically captures all terminal output (stdout and stderr) to a timestamped log file during setup execution. The log file is created under a dedicated `setup-logs` directory and is teed to the terminal so output remains visible in real time. Logging is skipped for help-related invocations.

# Files Changed

📄 `.setup/setup-common.sh`

Adds a new `setup_logging()` function that:
- Resolves a log root directory using `$BANK_OF_Z_WORK_DIR` (falling back to `$BANK_DIR`) and creates a `setup-logs` subdirectory if it does not already exist.
- Generates a timestamped log file name (e.g., `setup-20240101-120000.log`) to uniquely identify each run without per-line overhead.
- Redirects both stdout and stderr through `tee` so output is simultaneously written to the log file and displayed in the terminal.
- Prints an informational message indicating the active log file path.

The `main()` function is also updated to call `setup_logging()` immediately after detecting the execution mode, but only when a non-help phase is provided, ensuring that `-h`, `--help`, and `help` invocations are not affected.


.setup/setup-common.sh
```diff
@@ -543,6 +543,25 @@ print_usage() {
     echo "  IMS_DISABLED   Set to true to skip IMS-specific tests"
 }
 
+#########################################################
+# Logging setup
+#########################################################
+setup_logging() {
+    # Default log root: $BANK_OF_Z_WORK_DIR/setup-logs (falls back to $BANK_DIR/setup-logs)
+    local log_root="${BANK_OF_Z_WORK_DIR:-$BANK_DIR}/setup-logs"
```
chrishodgins: Do we need this fall-back? The `BANK_OF_Z_WORK_DIR` is set in `detect_bank_of_z_location` to the parent directory of `BANK_DIR`. If this is a grub setup, I'm not sure if the `setup-logs` directory would get deleted or not on the next sync.

Looking at my current `BANK_OF_Z_WORK_DIR` there is already a logs directory created by dbb and wazi-deploy. Maybe this should be set to `"${BANK_OF_Z_WORK_DIR:-$BANK_DIR}/logs/setup-boz"` or similar? 

We could consider defining a `sandbox.log_dir` variable in config.yaml and use that with `dbb_log_dir`/`deploy_log_dir` but that might be overkill.

.setup/setup-common.sh
```diff
@@ -543,6 +543,25 @@ print_usage() {
...
+    mkdir -p "$log_root"
+
+    local timestamp
+    timestamp=$(date '+%Y%m%d-%H%M%S')
+    LOG_FILE="${log_root}/setup-${timestamp}.log"
+
+    # Tee stdout+stderr to the log file while keeping terminal output intact.
+    # The timestamp is encoded in the log file name; no per-line overhead.
+    exec > >(tee -a "$LOG_FILE") 2>&1
```
chrishodgins: I suspect this doesn't interfere with the stdout/stderr redirection in the other child setup scripts as I doubt you would have any logs back but it might be worth a check if you haven't already.

chrishodgins (approved):
Change looks fine but I've left a few questions to consider.

